PR #5, opened 2023-05-22
Add SwiftLint and Unit Test Workflow

joelklabo:
- Add SwiftLint workflow
- Add Unit test workflow
- Add badges to README
- [x] Opt out of line_length violation

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.8
+// swift-tools-version: 5.7.1
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
--- a/Package.swift
+++ b/Package.swift
@@ -22,8 +22,10 @@ let package = Package(
         .target(
             name: "NostrSDK",
             dependencies: []
+        ),
         .testTarget(
             name: "NostrSDKTests",
             dependencies: ["NostrSDK"]
+        ),
     ]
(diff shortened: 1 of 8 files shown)


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.8
+// swift-tools-version: 5.7.1
```
tyiu, line 1: Oh, you might be using an older version than what I'm using.

Package.swift
```diff
@@ -10,23 +10,22 @@ let package = Package(
         // Products define the executables and libraries a package produces, and make them visible to other packages.
         .library(
             name: "NostrSDK",
-            targets: ["NostrSDK"]),
+            targets: ["NostrSDK"])
     ],
     dependencies: [
         // Dependencies declare other packages that this package depends on.
         // .package(url: /* package url */, from: "1.0.0"),
-        .package(url: "https://github.com/realm/SwiftLint.git", .upToNextMinor(from: "0.52.2"))
```
tyiu, line 18: I guess it makes sense that we don't want to necessarily pull in this dependency as part of what's required to use this library. We'll just have to run SwiftLint locally in a different way -- but I suppose the build at PR time should catch lint issues.

tyiu (approved):

bryanmontz (approved):

Issue #15, opened 2023-05-25
Create Structs and Parsing for Relay Message Types (Requests/Responses)


joelklabo:
### Client to Relay

| type    | description                                         | NIP         |
| ------- | --------------------------------------------------- | ----------- |
| `AUTH`  | used to send authentication events                  | [42](https://github.com/nostr-protocol/nips/blob/master/42.md) |
| `CLOSE` | used to stop previous subscriptions                 | [1](https://github.com/nostr-protocol/nips/blob/master/01.md)  |
| `COUNT` | used to request event counts                        | [45](https://github.com/nostr-protocol/nips/blob/master/45.md) |
| `EVENT` | used to publish events                              | [1](https://github.com/nostr-protocol/nips/blob/master/01.md)  |
| `REQ`   | used to request events and subscribe to new updates | [1](https://github.com/nostr-protocol/nips/blob/master/01.md)  |

### Relay to Client

| type     | description                                             | NIP         |
| -------- | ------------------------------------------------------- | ----------- |
| `AUTH`   | used to send authentication challenges                  | [42](https://github.com/nostr-protocol/nips/blob/master/42.md) |
| `COUNT`  | used to send requested event counts to clients          | [45](https://github.com/nostr-protocol/nips/blob/master/45.md) |
| `EOSE`   | used to notify clients all stored events have been sent | [1](https://github.com/nostr-protocol/nips/blob/master/01.md)  |
| `EVENT`  | used to send events requested to clients                | [1](https://github.com/nostr-protocol/nips/blob/master/01.md)  |
| `NOTICE` | used to send human-readable messages to clients         | [1](https://github.com/nostr-protocol/nips/blob/master/01.md)  |
| `OK`     | used to notify clients if an EVENT was successful       | [20](https://github.com/nostr-protocol/nips/blob/master/20.md) |

### Client to Relay (`RelayRequest`)

- [ ] `AUTH` // Requires signing
- [x] `CLOSE` #17 
- [x] `COUNT`#17
- [x] `EVENT` #17 
- [x] `REQ` #17 

### Relay to Client (`RelayResponse`)

- [x] `AUTH` #16 
- [x] `COUNT` #16 
- [x] `EOSE` #16 
- [x] `EVENT` #16 
- [x] `NOTICE` #16 
- [x] `OK` #16

joelklabo:
#16

PR #17, opened 2023-05-26
Relay Requests Encoding and Tests

joelklabo:
#15


Sources/NostrSDK/EventTag.swift
```diff
@@ -33,4 +33,19 @@ public struct EventTag: Codable, Equatable {
...
+
+    enum CodingKeys: CodingKey {
+        case identifier
+        case contentIdentifier
+        case recommendedRelayURL
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.unkeyedContainer()
+        try container.encode(self.identifier.rawValue)
```
bryanmontz, line 45: Please remove unnecessary `self.`s here.

Sources/NostrSDK/EventTag.swift
```diff
@@ -33,4 +33,19 @@ public struct EventTag: Codable, Equatable {
...
+        try container.encode(self.contentIdentifier)
+        if let url = self.recommendedRelayURL {
```
bryanmontz, line 47: There's a newer syntax:
```swift
if let recommendedRelayURL {
    ...
}
```

Sources/NostrSDK/Filter.swift
```diff
@@ -0,0 +1,56 @@
...
+        case kinds = "kinds"
+        case events = "#e"
+        case pubkeys = "#p"
+        case since = "since"
+        case until = "until"
+        case limit = "limit"
+    }
+}
+
+extension Filter: Codable {
```
bryanmontz, line 32: Did you need to implement the Codable requirements? Usually Swift can synthesize these for a simple struct.

Sources/NostrSDK/RelayRequest.swift
```diff
@@ -0,0 +1,40 @@
...
+        let payload = [AnyEncodable("EVENT"), AnyEncodable(event)]
+        return encode(payload: payload)
+    }
+
+    static func count(subscriptionId: String, filter: Filter) -> String? {
+        let payload = [AnyEncodable("COUNT"), AnyEncodable(subscriptionId), AnyEncodable(filter)]
+        return encode(payload: payload)
+    }
+
+    static func req(subscriptionId: String, filter: Filter) -> String? {
```
bryanmontz, line 29: Suggestion:
I'd prefer to use `request` here rather than `req` for easier readability. It's a technical implementation detail that it's actually "REQ" in the request.

Tests/NostrSDKTests/Helpers/JSONEquality.swift
```diff
@@ -0,0 +1,52 @@
+//
+//  JSONEquality.swift
+//  
+//
+//  Created by Joel Klabo on 5/26/23.
+//
+
+import Foundation
+
+func areEqualJSONArrayStrings(_ first: String?, _ second: String?) -> Bool {
```
bryanmontz, line 10: Test coverage needed on these. And they are declared globally. Maybe they could be in a protocol extension?

Tests/NostrSDKTests/RelayRequestEncodingTests.swift
```diff
@@ -0,0 +1,83 @@
...
+//  Created by Joel Klabo on 5/25/23.
+//
+
+@testable import NostrSDK
+import XCTest
+
+final class RelayRequestEncodingTests: XCTestCase, FixtureLoading {
+
+    func testEncodeClose() throws {
+        guard let request = RelayRequest.close(subscriptionId: "some-subscription-id") else {
```
bryanmontz, line 14: There's a better way to do this in tests:
```swift
let request = try XCTUnwrap(RelayRequest.close(subscriptionId: "some-subscription-id"), "failed to encode request")
```

Tests/NostrSDKTests/RelayRequestEncodingTests.swift
```diff
@@ -0,0 +1,83 @@
...
+        let eventTag = EventTag(identifier: .event, contentIdentifier: "93930d65435d49db723499335473920795e7f13c45600dcfad922135cf44bd63")
+        let pubkeyTag = EventTag(identifier: .pubkey, contentIdentifier: "f8e6c64342f1e052480630e27e1016dce35fc3a614e60434fef4aa2503328ca9")
+        let event = NostrEvent(id: "fa5ed84fc8eeb959fd39ad8e48388cfc33075991ef8e50064cfcecfd918bb91b",
+                               pubkey: "82341f882b6eabcd2ba7f1ef90aad961cf074af15b9ef44a09f9d2a8fbfbe6a2",
+                               createdAt: 1682080184,
+                               kind: .textNote,
+                               tags: [eventTag, pubkeyTag],
+                               content: "I think it stays persistent on your profile, but interface setting doesn’t persist. Bug.  ",
+                               signature: "96e6667348b2b1fc5f6e73e68fb1605f571ad044077dda62a35c15eb8290f2c4559935db461f8466df3dcf39bc2e11984c5344f65aabee4520dd6653d74cdc09")
+        guard let request = RelayRequest.event(event) else {
```
bryanmontz, line 34: Use `XCTUnwrap`. Recommended for all `guard`s used in test cases.

Sources/NostrSDK/Filter.swift
```diff
@@ -0,0 +1,56 @@
+//
+//  Filter.swift
+//  
+//
+//  Created by Joel Klabo on 5/26/23.
+//
+
+import Foundation
+
+struct Filter {
```
bryanmontz, line 10: I'm hoping that we might be able to have this raw filter private to the library, and expose a filter type to the client app that uses native types, such as a `NostrKind` enum for the `kinds` part of the filter and `Keypair` types for the `pubkeys`. It could be converted the raw type when making requests, but that can come later.

Sources/NostrSDK/EventTag.swift
joelklabo: 7ce444d

Sources/NostrSDK/EventTag.swift
joelklabo: 7ce444d

Sources/NostrSDK/Filter.swift
joelklabo: Good catch, you're right I don't: 6fa6eb7

Sources/NostrSDK/RelayRequest.swift
joelklabo: fixed: 6cd025c

Tests/NostrSDKTests/RelayRequestEncodingTests.swift
joelklabo: Yeah, that's much nicer 🤙 fixed: fa1b42e

Tests/NostrSDKTests/Helpers/JSONEquality.swift
joelklabo: Yeah, much better. Fixed: 5f8947f

Sources/NostrSDK/Filter.swift
joelklabo: What do you think we should for this right now? I think when we start setting up the public API we should definitely do that but not sure if we need to now. What do you think?

bryanmontz (approved):

PR #24, opened 2023-05-28
Add dependency to secp256k1 and add structs for keys

tyiu:


Sources/NostrSDK/Keys.swift
```diff
@@ -0,0 +1,134 @@
...
+import Foundation
+import secp256k1
+
+struct Keypair {
+    public let publicKey: PublicKey
+    public let privateKey: PrivateKey
+
+    init?() {
+        guard let key = try? secp256k1.Signing.PrivateKey() else {
+            print("Could not generate new secp256k1 private key.")
```
tyiu, line 17: Maybe it makes sense to not swallow any of the errors in this file but re-throw? What do you think?

Package.swift
```diff
@@ -5,7 +5,6 @@ import PackageDescription
 
 let package = Package(
     name: "NostrSDK",
-    platforms: [.macOS(.v12)],
```
tyiu, line 8: This came from pulling in SwiftLint initially, which got pulled out, but we forgot to remove this line.

Tests/NostrSDKTests/Bech32Tests.swift
```diff
@@ -0,0 +1,153 @@
...
+            } catch let error as Bech32.DecodingError {
+                XCTAssert(errorsEqual(error, reason), "Decoding error mismatch, got \(error.localizedDescription), expected \(reason.localizedDescription)")
+            } catch {
+                XCTFail("Invalid error occured: \(error.localizedDescription)")
+            }
+        }
+    }
+
+    private func errorsEqual(_ lhs: Bech32.DecodingError, _ rhs: Bech32.DecodingError) -> Bool {
+        switch lhs {
```
bryanmontz, line 127: There's a nicer way to do this using a tuple in the `switch`:
```swift
switch (lhs, rhs) {
case (.checksumMismatch, .checksumMismatch),
         (.incorrectChecksumSize, .incorrectChecksumSize),
        ...:
        return true 
default:
        return false
}
```

Tests/NostrSDKTests/Bech32Tests.swift
```diff
@@ -0,0 +1,153 @@
...
+        case .nonUTF8String:
+            return rhs == .nonUTF8String
+        case .stringLengthExceeded:
+            return rhs == .stringLengthExceeded
+        case .nonPrintableCharacter:
+            return rhs == .nonPrintableCharacter
+        }
+    }
+
+    static var allTests = [
```
bryanmontz, line 149: Remove unused code

Sources/NostrSDK/Keys.swift
```diff
@@ -0,0 +1,134 @@
...
+//  Keys.swift
+//  
+//
+//  Created by Terry Yiu on 5/23/23.
+//
+
+import Foundation
+import secp256k1
+
+struct Keypair {
```
bryanmontz, line 11: I assume we'll want to make these `struct`s and all `init`s `public`.

Sources/NostrSDK/Keys.swift
bryanmontz: If this is the kind of thing that never actually fails, maybe it's unnecessary to burden the call sites.

Tests/NostrSDKTests/Bech32Tests.swift
```diff
@@ -0,0 +1,153 @@
+//
+//  Bech32Tests.swift
+//
+//  Modified by Terry Yiu on 5/24/23.
+//  Created by Evolution Group Ltd on 12.02.2018.
+//  Copyright © 2018 Evolution Group Ltd. All rights reserved.
```
bryanmontz, line 6: Are you sure this is okay to copy in to our project with an MIT license? I didn't check. Just wanting to make sure we do due diligence.

Tests/NostrSDKTests/Bech32Tests.swift
```diff
@@ -0,0 +1,153 @@
...
+        for invalid in _invalidChecksum {
+            let checksum = invalid.bech32
+            let reason = invalid.error
+            do {
+                let decoded = try Bech32.decode(checksum)
+                XCTFail("Successfully decoded an invalid checksum \(checksum): \(decoded.checksum.hexString)")
+            } catch let error as Bech32.DecodingError {
+                XCTAssert(errorsEqual(error, reason), "Decoding error mismatch, got \(error.localizedDescription), expected \(reason.localizedDescription)")
+            } catch {
+                XCTFail("Invalid error occured: \(error.localizedDescription)")
```
bryanmontz, line 121: Spelling: "occured" -> "occurred"

Tests/NostrSDKTests/Bech32Tests.swift
tyiu: I sourced it from here and it's also MIT licensed. https://github.com/0xDEADP00L/Bech32

I should ask the folks from Nos, though, for some of the other code.

tyiu: Permission to redistribute this code under MIT license granted nostr:note1q39598qkdc093sdq4enudjf0dall76s7n779k07nutgd9r2zt6vq96l8c2

joelklabo (approved):

Package.swift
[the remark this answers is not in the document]
tyiu: I assume it supports everything by default, and you use these to scope it down.

Sources/NostrSDK/Data+Encoding.swift
```diff
@@ -0,0 +1,71 @@
...
+//
+//
+//  Created by Terry Yiu on 5/28/23.
+//
+
+import Foundation
+
+extension Data {
+    var hexString: String {
+        let hexDigits = Array("0123456789abcdef".utf16)
```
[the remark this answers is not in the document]
tyiu: I copied this from Nos so I don't want to touch it as it works.

Tests/NostrSDKTests/Bech32Tests.swift
```diff
@@ -0,0 +1,153 @@
...
+import XCTest
+@testable import NostrSDK
+
+private typealias InvalidChecksum = (bech32: String, error: Bech32.DecodingError)
+private typealias ValidAddressData = (address: String, script: [UInt8])
+private typealias InvalidAddressData = (hrp: String, version: Int, programLen: Int)
+
+class Bech32Tests: XCTestCase {
+
+    private let _validChecksum: [String] = [
```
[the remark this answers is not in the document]
tyiu: From @bryanmontz 

>No reason to start a property name with an underscore in Swift. I assumed it was code copied from somewhere else. In which case I’d probably just leave it.

Tests/NostrSDKTests/Bech32Tests.swift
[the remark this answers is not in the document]
tyiu: Came from the copied file. This test file was testing a few other things that we don't need. I deleted them but forgot to clean this up.

PR #25, opened 2023-05-29
simplify decodeFixture

bryanmontz:
This PR simplifies `decodeFixture`. Since we are already in a throwing context, there's no need to use `do-catch`. And we can pass the generic type in by specifying the type of the return parameter at the call site rather than as a parameter into the function.


joelklabo:
Awesome 👏

joelklabo (approved):

PR #29, opened 2023-06-10
add Relay, provides the communication interface to relays

bryanmontz:
This PR adds the communication interface to relays.

* Connect to a relay
* Create a subscription based on a provided `Filter`
* Easily listen for and respond to events and state changes

### DIFF
--- a/Sources/NostrSDK/Filter.swift
+++ b/Sources/NostrSDK/Filter.swift
@@ -28,6 +28,11 @@ public struct Filter: Codable {
         case limit = "limit"
     }
     
+    /// Creates the filter with the specified parameters
+    ///
+    /// - Parameters:
+    ///   - ids: a list of event ids or prefixes
+    ///   - authors: a list of pubkeys or prefixes, the pubkey of an event must be one of these
     public init(ids: [String]? = nil, authors: [String]? = nil, kinds: [Int]? = nil, events: [String]? = nil, pubkeys: [String]? = nil, since: Int? = nil, until: Int? = nil, limit: Int? = nil) {
         self.ids = ids
         self.authors = authors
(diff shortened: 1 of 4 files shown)


Sources/NostrSDK/Filter.swift
```diff
@@ -28,6 +28,11 @@ public struct Filter: Codable {
         case limit = "limit"
     }
     
+    /// Creates the filter with the specified parameters
+    ///
+    /// - Parameters:
+    ///   - ids: a list of event ids or prefixes
```
tyiu, line 34: It's clear that the ids and authors should be the hexadecimal representations when you read https://github.com/nostr-protocol/nips/blob/master/01.md but I think it would be valuable to also mention it in our public docs. Otherwise, some may accidentally pass in https://github.com/nostr-protocol/nips/blob/master/19.md bech32 encoded identifiers. Non-blocking, but further down, it would be interesting to explore what kind of validations we want in place to just prevent that entire class of problems.

Tests/NostrSDKTests/RelayTests.swift
```diff
@@ -0,0 +1,48 @@
...
+@testable import NostrSDK
+import XCTest
+
+final class RelayTests: XCTestCase {
+    
+    private static let RelayURL = URL(string: "wss://relay.damus.io")!
+    
+    private var cancellables = Set<AnyCancellable>()
+
+    func testConnectAndReceive() async throws {
```
tyiu, line 18: Spoke offline about this with Bryan. This test will do a live connection with the Damus relay, which is fine for now just to verify that this change works. As a future iteration, it would be great to spin up a local in-memory relay in tests (not necessarily in the public SDK interface) that accepts websocket connections and naively returns basic responses.

tyiu (approved):

PR #41, opened 2023-06-14
add NIP-05 validation

bryanmontz:
Closes https://github.com/nostr-sdk/nostr-sdk-ios/issues/33

### DIFF
--- a/Sources/NostrSDK/NIP05Validating.swift
+++ b/Sources/NostrSDK/NIP05Validating.swift
@@ -45,12 +45,6 @@ struct NIP05Response: Codable {
     let relays: [String: [String]]?
 }
 
-public protocol DataRequesting {
-    func data(from url: URL, delegate: URLSessionTaskDelegate?) async throws -> (Data, URLResponse)
-}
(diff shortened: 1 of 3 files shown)


Sources/NostrSDK/NIP05Validating.swift
```diff
@@ -0,0 +1,157 @@
...
+    /// Use this to look up the pubkey associated with an email-like [internet identifier](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1).
+    let names: [String: String]?
+    
+    /// An object with public keys as properties and arrays of relay URLs as values
+    ///
+    /// Use this to learn in which relays a specific user may be found.
+    let relays: [String: [String]]?
+}
+
+public protocol DataRequesting {
```
tyiu, line 48: Should we refactor this and other generic HTTP request response stuff out? I imagine there could be other NIPs that might use it.

tyiu (approved):
Looks great. Left a comment about extracting some bits of code into its own file.

PR #43, opened 2023-06-19
Add finer-grained relay connection functions

tyiu:


Sources/NostrSDK/Relay.swift
```diff
@@ -148,6 +145,25 @@ public final class Relay: ObservableObject {
...
+        guard state != .connected && state != .connecting else {
+            return
+        }
+
+        state = .connecting
+        socket.connect()
+    }
+
+    /// Disconnects from the relay if it is in a connected or connecting state.
+    public func disconnect() {
```
tyiu, line 160: I'm having difficulty testing this function. If I add a `disconnect()` call to the end of each of the `RelayTests` functions, it seems to usually break `testRelayDelegate` and I don't understand why yet.

bryanmontz: Added a commit to fix. You can't call `fulfill()` more than once on an `XCTestExpectation` and the relay was sending the state to the delegate even when the state didn't change, so it would fulfill the same expectation more than once and crash sometimes. I fixed that and added a disconnect expectation. That's better coverage! Thanks for thinking about testing disconnect.

PR #47, opened 2023-07-11
Add NIP-25 reactions

tyiu:
https://github.com/nostr-protocol/nips/blob/master/25.md

This implementation is missing custom emoji reactions, but I want to handle that in a separate PR because custom emojis can be used in other places outside of reactions.
https://github.com/nostr-protocol/nips/blob/master/25.md#custom-emoji-reaction

### DIFF
--- a/Sources/NostrSDK/Events/ReactionEvent.swift
+++ b/Sources/NostrSDK/Events/ReactionEvent.swift
@@ -1,5 +1,5 @@
 //
-//  File.swift
+//  ReactionEvent.swift
 //  
 //
 //  Created by Terry Yiu on 8/12/23.
--- a/Sources/NostrSDK/Events/ReactionEvent.swift
+++ b/Sources/NostrSDK/Events/ReactionEvent.swift
@@ -1,0 +1,21 @@
+//
+//  File.swift
+//  
+//
+//  Created by Terry Yiu on 8/12/23.
+//
+
+import Foundation
+
+/// A reaction event (kind 7) in response to a different event.
+///
+/// See [NIP-25](https://github.com/nostr-protocol/nips/blob/master/25.md).
+public class ReactionEvent: NostrEvent {
+    public var reactedEventId: String? {
+        tags.last(where: { $0.name == .event })?.value
+    }
+
+    public var reactedEventPubkey: String? {
+        tags.last(where: { $0.name == .pubkey })?.value
+    }
+}
--- a/Tests/NostrSDKTests/EventCreatingTests.swift
+++ b/Tests/NostrSDKTests/EventCreatingTests.swift
@@ -6,7 +6,7 @@
 //
 
 import Foundation
-import NostrSDK
+@testable import NostrSDK
 import XCTest
 
 final class EventCreatingTests: XCTestCase, EventCreating, EventVerifying {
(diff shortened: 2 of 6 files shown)


bryanmontz (approved):

bryanmontz:
This should probably include the new subclassing too, so I might add to it before merging.

Tests/NostrSDKTests/EventCreatingTests.swift
```diff
@@ -61,4 +61,24 @@ final class EventCreatingTests: XCTestCase, EventCreating, EventVerifying {
         XCTAssertThrowsError(try recommendServerEvent(withRelayURL: inputURL,
                                                       signedBy: Keypair.test))
     }
+
+    func testCreateReactionEvent() throws {
+        let reactedEvent = try textNote(withContent: "Hello world!",
```
tyiu, line 66: A more thorough test would be to react to a threaded reply text note to ensure that all the `e` and `p` tags (and nothing else) get copied over to the reaction event, but we currently don't handle replies yet. I'll add this test case in after we get better support for replies.

Tests/NostrSDKTests/Fixtures/reaction.json
```diff
@@ -0,0 +1,19 @@
...
+    "created_at": 1689029084,
+    "id": "8a3217770794fabe89adac500dcd5d38966d3ba3cb83fabc97b58135980f76cd",
+    "kind": 7,
+    "pubkey": "2779f3d9f42c7dee17f0e6bcdcf89a8f9d592d19e3b1bbd27ef1cffd1a7f98d1",
+    "sig": "c0dea5d4612d834e13e0dcfeff71a345f761d868bf27fd5e3fe521b76872d5da3db05375f8739a4bad86189d63720187c08170827990b113b477437f17e4a906",
+    "tags":
+    [
+        [
+            "e",
+            "62dcc905c282dd712bbe6b47d2e40feb333f8a0c39899617f4ca37337199ede0"
```
tyiu, line 12: On an unrelated note, if I remove this value, we get a crash. We should be more resilient against malicious or poorly coded clients adding "malformed" tags.

Tests/NostrSDKTests/EventCreatingTests.swift
```diff
@@ -61,4 +61,26 @@ final class EventCreatingTests: XCTestCase, EventCreating, EventVerifying {
...
+        XCTAssertEqual(event.reactedEventPubkey, reactedEvent.pubkey)
+        XCTAssertEqual(event.content, "🤙")
+
+        let expectedTags = [
+            Tag(name: .event, value: reactedEvent.id),
+            Tag(name: .pubkey, value: reactedEvent.pubkey)
+        ]
+        XCTAssertEqual(event.tags, expectedTags)
+
+        try verifyEvent(event)
```
tyiu, line 84: Amended my commit to include this line to verify the event signature.

Tests/NostrSDKTests/Fixtures/reaction.json
joelklabo: Do you know which test fails?

tyiu: The event decoding ones. The Tag initializer requires there to be a tag name and value.

Sources/NostrSDK/Events/ReactionEvent.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  File.swift
```
bryanmontz, line 2: It's annoying that this is here by default, but let's update it I guess.

Sources/NostrSDK/Events/ReactionEvent.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+
+import Foundation
+
+/// A reaction event (kind 7) in response to a different event.
+///
+/// See [NIP-25](https://github.com/nostr-protocol/nips/blob/master/25.md).
+public class ReactionEvent: NostrEvent {
+    public var reactedEventId: String? {
+        tags.last(where: { $0.name == .event })?.value
```
bryanmontz, line 15: Why `last` on these? Is that intentional (as opposed to `first`)?

Tests/NostrSDKTests/Fixtures/reaction.json
bryanmontz: Absolutely. I haven't even started to try to think of all the ways these could go wrong. But yeah we definitely should add tests for all kinds of nonsense we could see from relays or clients.

Sources/NostrSDK/Events/ReactionEvent.swift
tyiu: It's intentional per the spec.

Sources/NostrSDK/Events/ReactionEvent.swift
tyiu: Good catch. I'll update it before merging.

Sources/NostrSDK/Events/ReactionEvent.swift
tyiu: https://github.com/nostr-protocol/nips/blob/master/25.md#tags

```
The last e tag MUST be the id of the note that is being reacted to.
The last p tag MUST be the pubkey of the event being reacted to.
```

Tests/NostrSDKTests/Fixtures/reaction.json
tyiu: https://github.com/nostr-sdk/nostr-sdk-ios/issues/66

PR #55, opened 2023-07-18
Expand EventKind enum to support unknown event kind numbers

tyiu:
Inspired by this StackOverflow answer: https://stackoverflow.com/a/61150450


bryanmontz (approved):
Very cool. And yeah, that's a great SO post. I've never used `RawRepresentable` before. Feels weird to pull the Int values off of the case declarations, but they're in the `rawValue` implementation.

PR #56, opened 2023-07-23
establish NostrEvent subclassing pattern with basic event kinds

bryanmontz:
This PR implements a class hierarchy architecture pattern for the parent event type `NostrEvent` and all of its many specific kinds.

The general pattern is that kinds will map to specific subclasses, which will then interpret the "raw" information contained in the parent class, primarily from the the `content` and `tags` properties.

Closes https://github.com/nostr-sdk/nostr-sdk-ios/issues/48
Closes https://github.com/nostr-sdk/nostr-sdk-ios/issues/49
Closes https://github.com/nostr-sdk/nostr-sdk-ios/issues/50
Closes https://github.com/nostr-sdk/nostr-sdk-ios/issues/51

### DIFF
--- a/Sources/NostrSDK/Events/RecommendServerEvent.swift
+++ b/Sources/NostrSDK/Events/RecommendServerEvent.swift
@@ -1,0 +1,17 @@
+//
+//  RecommendServerEvent.swift
+//  
+//
+//  Created by Bryan Montz on 7/23/23.
+//
+
+import Foundation
+
+/// An event that contains a relay the event creator wants to recommend to its followers.
+///
+/// > Note: [NIP-01 Specification](https://github.com/nostr-protocol/nips/blob/b503f8a92b22be3037b8115fe3e644865a4fa155/01.md#basic-event-kinds)
+final class RecommendServerEvent: NostrEvent {
+    var relayURL: URL? {
+        URL(string: content)
+    }
+}
--- a/Sources/NostrSDK/Events/SetMetadataEvent.swift
+++ b/Sources/NostrSDK/Events/SetMetadataEvent.swift
@@ -1,0 +1,62 @@
+//
+//  SetMetadataEvent.swift
+//  
+//
+//  Created by Bryan Montz on 7/22/23.
+//
+
+import Foundation
+
+/// An object that describes a user.
+public struct UserMetadata: Decodable {
+    
+    /// The user's name.
+    public let name: String?
+    
+    /// The user's description of themself.
+    public let about: String?
+    
+    /// The user's website address.
+    public let website: URL?
+    
+    /// The user's Nostr address.
+    ///
+    /// > Note: [NIP-05 Specification](https://github.com/nostr-protocol/nips/blob/master/05.md#nip-05).
+    public let nostrAddress: String?
+    
+    /// A URL to retrieve the user's picture.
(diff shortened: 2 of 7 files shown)


Sources/NostrSDK/Events/RecommendServerEvent.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Bryan Montz on 7/23/23.
+//
+
+import Foundation
+
+/// An event that contains a relay the event creator wants to recommend to its followers.
+///
+/// > Note: [NIP-01 Specification](https://github.com/nostr-protocol/nips/blob/b503f8a92b22be3037b8115fe3e644865a4fa155/01.md#basic-event-kinds)
+final class RecommendServerEvent: NostrEvent {
+    var relayURL: URL? {
```
tyiu, line 14: Would there be any validation we want to perform here to ensure it's a websocket URL (i.e., `ws://` or `wss://`) and return `nil` if not?

Sources/NostrSDK/Events/SetMetadataEvent.swift
```diff
@@ -0,0 +1,62 @@
...
+public struct UserMetadata: Decodable {
+    
+    /// The user's name.
+    public let name: String?
+    
+    /// The user's description of themself.
+    public let about: String?
+    
+    /// The user's website address.
+    public let website: URL?
```
tyiu, line 20: Oh, interesting. The decoder is smart enough to transform strings into URLs and vice-versa?

tyiu (approved):
Hype! This is great.

Sources/NostrSDK/Events/SetMetadataEvent.swift
bryanmontz: This surprised me too, but yeah!

Sources/NostrSDK/Events/RecommendServerEvent.swift
bryanmontz: Not sure. We could if we want because the spec specifically says it's for relays and those will be websockets for now.

tyiu:
One more thought: this change adds support for decoding -- what are your thoughts about encoding subclass instances for signing for submission to relays?

Sources/NostrSDK/Events/RecommendServerEvent.swift
tyiu: For this specific case, I think we can be opinionated here because any other non-websocket protocol URL does not make on this field.

bryanmontz:
I don't think the subclasses will need to worry about signing at all. The base event type can handle that after simple, specific-kind initializers pack all the right fields of an event. I'll tackle that for these first kinds next.

tyiu:
> I don't think the subclasses will need to worry about signing at all. The base event type can handle that after simple, specific-kind initializers pack all the right fields of an event. I'll tackle that for these first kinds next.

Sorry, I didn't mean that the subclasses would be responsible for signing. I just meant that the initialization of the subclass instance with the kind specific fields would/should also call the base NostrEvent class constructor initializer (so that when event signing does happen, it can just rely on the base NostrEvent fields when encoding and signing the message). Basically, what you said in the latter part.

joelklabo (approved):

PR #64, opened 2023-08-12
Refactor Demo App

joelklabo:
Updated the demo app to organize the views better and be able to share a relay connection:

Simulator Screen Recording - iPhone 14 Pro - 2023-08-12 at 11 04 49

### DIFF
--- a/demo/NostrSDKDemo/Views/ConnectRelayView.swift
+++ b/demo/NostrSDKDemo/Views/ConnectRelayView.swift
@@ -79,9 +79,8 @@ struct ConnectRelayView: View {
 
 struct ConnectRelayView_Previews: PreviewProvider {
     static var previews: some View {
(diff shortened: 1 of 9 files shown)


demo/NostrSDKDemo/Views/ConnectRelayView.swift
```diff
@@ -11,20 +11,28 @@ import Combine
 
 struct ConnectRelayView: View {
 
+    @Binding var relay: Relay?
```
bryanmontz, line 14: Why do we need a `Binding` to a relay rather than just a relay?

joelklabo: I am passing it in from the `ContentView` so each view that needs it can share the relay. But, maybe that's not necessary for it to be a `@Binding`. I basically just default to that when I pass a `@State` to another object.

bryanmontz (approved):

Issue #67, opened 2023-08-13
Add a default relay or list of relays to select from to Demo App


bryanmontz:
It's a pain to have to type in a relay URL every time you run the app.

PR #69, opened 2023-08-13
Auto-Connect to Relay on Start

joelklabo:
#67


bryanmontz (approved):
This will help. Thanks! Maybe we can add a picker with several options later.

joelklabo:
> This will help. Thanks! Maybe we can add a picker with several options later.

Yeah, that's a good idea

PR #71, opened 2023-08-14
Add Demo of Sending Direct Message

joelklabo:
IMG_3909
Simulator Screenshot - iPhone 14 Pro - 2023-08-13 at 17 21 23

### DIFF
--- a/demo/NostrSDKDemo/Demo Views/DirectMessageDemoView.swift
+++ b/demo/NostrSDKDemo/Demo Views/DirectMessageDemoView.swift
@@ -1,0 +1,66 @@
+//
+//  DirectMessageDemoView.swift
+//  NostrSDKDemo
+//
+//  Created by Honk on 8/13/23.
+//
+
+import SwiftUI
+import NostrSDK
+
+struct DirectMessageDemoView: View, EventCreating {
+
+    @Binding var relay: Relay?
+
+    @State private var recipientPublicKey = ""
+    @State private var recipientPublicKeyIsValid: Bool = false
+
+    @State private var senderPublicKey = ""
+    @State private var senderPublicKeyIsValid: Bool = false
+
+    @State private var senderPrivateKey = ""
+    @State private var senderPrivateKeyIsValid: Bool = false
+
+    @State private var message: String = ""
+
+    var body: some View {
+        RelayFormView(relay: $relay) {
+            Section("Recipient") {
+                KeyInputSectionView(key: $recipientPublicKey,
+                                    isValid: $recipientPublicKeyIsValid,
+                                    type: .public)
+            }
+            Section("Sender") {
+                KeyInputSectionView(key: $senderPublicKey,
+                                    isValid: $senderPublicKeyIsValid,
+                                    type: .public)
+                KeyInputSectionView(key: $senderPrivateKey,
+                                    isValid: $senderPrivateKeyIsValid,
+                                    type: .private)
+            }
+            Section("Message") {
+                TextField("Enter a message.", text: $message)
+            }
+            Button("Send") {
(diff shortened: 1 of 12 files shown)


demo/NostrSDKDemo/Demo Views/DirectMessageDemoView.swift
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    private func readyToSend() -> Bool {
+        if !message.isEmpty &&
+            recipientPublicKeyIsValid &&
+            senderPrivateKeyIsValid {
+            return true
+        }
+        return false
```
bryanmontz, line 81: You don't need the returns:
```suggestion
        !message.isEmpty &&
        recipientPublicKeyIsValid &&
        senderPrivateKeyIsValid
```

demo/NostrSDKDemo/Views/KeyInputSectionView.swift
```diff
@@ -0,0 +1,135 @@
...
+        }
+        .onChange(of: key) { _ in
+            isValid = checkValidity(key: key, type: type)
+        }
+        .onAppear {
+            isValid = checkValidity(key: key, type: type)
+        }
+    }
+
+    private func label(for type: KeyType) -> String {
```
bryanmontz, line 49: Add this to the `KeyType` enum itself like this:
```swift
var label: String {
    switch self {
    case .public: return "Public Key"
    case .private: return "Private Key"
    }
}
```

Then the call sites are much cleaner:
```swift
keyType.label
```

demo/NostrSDKDemo/Views/KeyInputSectionView.swift
```diff
@@ -0,0 +1,135 @@
...
+            return checkValidity(publicKey: key)
+        case .private:
+            return checkValidity(privateKey: key)
+        }
+    }
+
+    // swiftlint:disable unused_optional_binding
+    private func checkValidity(publicKey: String) -> Bool {
+        if publicKey.contains("npub") {
+            if let _ = PublicKey(npub: publicKey) {
```
bryanmontz, line 70: These 5 lines reduce to:
```swift
return PublicKey(pub: publicKey) != nil
```

Can be applied to all the other cases too.

demo/NostrSDKDemo/Views/KeyInputSectionView.swift
```diff
@@ -0,0 +1,135 @@
...
+    private func label(for type: KeyType) -> String {
+        switch type {
+        case .public:
+            return "Public Key"
+        case .private:
+            return "Private Key"
+        }
+    }
+
+    private func checkValidity(key: String, type: KeyType) -> Bool {
```
bryanmontz, line 58: How about:
```swift
private func isKeyValid(_ key: String, type: KeyType) -> Bool
```

bryanmontz (changes requested):
Good work, Joel! Few small comments.

demo/NostrSDKDemo/Views/KeyInputSectionView.swift
joelklabo: 🤙

demo/NostrSDKDemo/Views/KeyInputSectionView.swift
joelklabo: Nice, was able to get rid of this too: `    // swiftlint:disable unused_optional_binding
`

demo/NostrSDKDemo/Views/KeyInputSectionView.swift
joelklabo: 👍

PR #75, opened 2023-08-19
add contact list creation and tests

bryanmontz:
Adds the ability to create contact lists from a list of pubkeys or pubkey tags. I provided both options so it can be very simple if the client prefers, but it also has full functionality matching the spec (for petnames).

### DIFF
--- a/Sources/NostrSDK/Events/ContactListEvent.swift
+++ b/Sources/NostrSDK/Events/ContactListEvent.swift
@@ -32,8 +32,17 @@ public struct RelayPermissions: Equatable {
 /// > Note: [NIP-02 Specification](https://github.com/nostr-protocol/nips/blob/master/02.md#contact-list-and-petnames)
 public final class ContactListEvent: NostrEvent {
     
-    convenience init(tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
-        try self.init(kind: .contactList, content: "", tags: tags, createdAt: createdAt, signedBy: keypair)
+    public required init(from decoder: Decoder) throws {
+        try super.init(from: decoder)
+    }
+    
+    @available(*, unavailable, message: "This initializer is unavailable for this class.")
+    override init(kind: EventKind, content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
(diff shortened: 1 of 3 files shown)


Sources/NostrSDK/Events/ContactListEvent.swift
```diff
@@ -32,6 +32,19 @@ public struct RelayPermissions: Equatable {
 /// > Note: [NIP-02 Specification](https://github.com/nostr-protocol/nips/blob/master/02.md#contact-list-and-petnames)
 public final class ContactListEvent: NostrEvent {
     
+    public required init(from decoder: Decoder) throws {
+        try super.init(from: decoder)
+    }
+    
+    @available(*, unavailable, message: "This initializer is unavailable for this class.")
+    override init(kind: EventKind, content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
```
bryanmontz, line 40: I'm trying something fairly opinionated here, so please let me know your feedback. I'm marking the parent class's initializer as unavailable for this class. That way I can prevent client apps from creating an event like this, where the class and the kind don't match:
```swift
let badEvent = ContactListEvent(kind: .recommendServer, tags:...
```

Another great thing about this is that code completion won't even suggest that initializer.

joelklabo: Yeah, I think that's a good idea. You can always use the `NostrEvent` initializer if needed for something custom.

joelklabo (approved):
Only suggestion is to verify the note in the test. Only because that's what we've done for the others. But, I could definitely be persuaded that it isn't necessary since you're not adding anything that has an effect on verification.

Tests/NostrSDKTests/EventCreatingTests.swift
```diff
@@ -81,6 +61,58 @@ final class EventCreatingTests: XCTestCase, EventCreating, EventVerifying {
...
+        let tags: [Tag] = [
+            Tag(name: .pubkey, value: "83y9iuhw9u0t8thw8w80u", otherParameters: ["bob"]),
+            Tag(name: .pubkey, value: "19048ut34h23y89jio3r8", otherParameters: ["alice"]),
+            Tag(name: .pubkey, value: "5r623gyewfbh8uuiq83rd", otherParameters: ["steve"])
+        ]
+        
+        let event = try contactList(withPubkeyTags: tags,
+                                    signedBy: Keypair.test)
+        
+        XCTAssertEqual(event.tags, tags)
```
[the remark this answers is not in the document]
bryanmontz: I just forgot. We should always do that. Thanks!

PR #78, opened 2023-08-20
preclude NostrEvent subclasses from using the initializer that allows specifying the kind explicitly

bryanmontz:
Implements the idea from https://github.com/nostr-sdk/nostr-sdk-ios/pull/75 where we want to prevent disparities between the `NostrEvent` subclass and it's `kind` property.


joelklabo (approved):
Nice I think this is a good improvement

PR #79, opened 2023-08-20
add ability to create reposts and tests

bryanmontz:


Sources/NostrSDK/Events/TextNoteRepostEvent.swift
```diff
@@ -12,6 +12,10 @@ import Foundation
 /// See [NIP-18](https://github.com/nostr-protocol/nips/blob/master/18.md#reposts).
 public final class TextNoteRepostEvent: GenericRepostEvent {
     
+    override class var kind: EventKind {
+        .repost
```
joelklabo, line 16: Curious about this, it looks to me like it's overriding the value from `GenericRepostEvent` but they are the same

bryanmontz: Dang, you're right! Good catch. Fixed and added a test that would've caught it.

joelklabo: Nice! 👍

joelklabo (approved):

PR #82, opened 2023-08-21
Parse display_name

joelklabo:
I don't think this is in the spec but it is really common and the default in Damus for example so I think we should have it available if it's there.

### DIFF
--- a/Sources/NostrSDK/Events/SetMetadataEvent.swift
+++ b/Sources/NostrSDK/Events/SetMetadataEvent.swift
@@ -13,6 +13,9 @@ public struct UserMetadata: Codable {
     /// The user's name.
     public let name: String?
     
+    /// The user's display name.
+    public let displayName: String?
+
     /// The user's description of themself.
     public let about: String?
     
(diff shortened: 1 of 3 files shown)


Sources/NostrSDK/Events/SetMetadataEvent.swift
```diff
@@ -13,6 +13,9 @@ public struct UserMetadata: Codable {
     /// The user's name.
     public let name: String?
     
+    /// The user's display name.
```
bryanmontz, line 16: I think supporting out-of-spec attributes is something we just have to accept as part of an open protocol, but I would like us to note that fact in the DocC comments whenever we do this. That way, client developers will be cautious about depending on such properties.

```suggestion
    /// The user's display name.
    /// > Warning: This property is not part of the Nostr specifications.
```

You could also use `Note` if you feel that `Warning` is too strong.

joelklabo: I think warning is OK for now, if someone goes to look at that documentation it should be clear that it is not supported.

bryanmontz (approved):

PR #86, opened 2023-08-25
Send RelayEvent Containing NostrEvent and Subscription ID

joelklabo:


Sources/NostrSDK/Relay.swift
```diff
@@ -133,9 +139,10 @@ public final class Relay: ObservableObject, EventVerifying {
         func handle(messageData: Data) {
             let response = RelayResponse.decode(data: messageData)
             switch response {
-            case .event(_, let event):
-                events.send(event)
-                delegate?.relay(self, didReceive: event)
+            case .event(let subscriptionId, let event):
+                let relayEvent = RelayEvent(event: event, subscriptionId: subscriptionId)
+                events.send(relayEvent)
```
joelklabo, line 144: For the publisher case, maybe a tuple would be better...

tyiu (approved):

PR #99, opened 2023-09-30
add support for relay OK messages

bryanmontz:
defined in NIP-01

Closes https://github.com/nostr-sdk/nostr-sdk-ios/issues/85


tyiu (approved):
Looks good and seems to conform to NIP-01.

PR #104, opened 2023-11-12
Add support for NIP-30 Custom Emojis

tyiu:
https://github.com/nostr-sdk/nostr-sdk-ios/issues/103

### DIFF
--- a/Sources/NostrSDK/CustomEmoji.swift
+++ b/Sources/NostrSDK/CustomEmoji.swift
@@ -16,12 +16,12 @@ public class CustomEmoji: CustomEmojiValidating, Equatable {
     /// A name given for the emoji, which MUST be comprised of only alphanumeric characters and underscores.
     let shortcode: String
 
-    /// A a URL to the corresponding image file of the emoji.
-    let imageUrl: URL
+    /// A URL to the corresponding image file of the emoji.
+    let imageURL: URL
 
-    init?(shortcode: String, imageUrl: URL) {
+    init?(shortcode: String, imageURL: URL) {
         self.shortcode = shortcode
-        self.imageUrl = imageUrl
+        self.imageURL = imageURL
 
         if !isValidShortcode(shortcode: shortcode) {
             return nil
@@ -29,12 +29,12 @@ public class CustomEmoji: CustomEmojiValidating, Equatable {
     }
 
     func tag() -> Tag {
-        Tag(name: .emoji, value: shortcode, otherParameters: [imageUrl.absoluteString])
+        Tag(name: .emoji, value: shortcode, otherParameters: [imageURL.absoluteString])
     }
 
     func isEqual(to customEmoji: CustomEmoji) -> Bool {
         shortcode == customEmoji.shortcode &&
-        imageUrl == customEmoji.imageUrl
+        imageURL == customEmoji.imageURL
     }
 }
 
@@ -44,11 +44,11 @@ public extension CustomEmojiInterpreting {
     /// It does not fetch images from the specified URLs to determine if they exist.
     var customEmojis: [CustomEmoji] {
         return tags.compactMap { tag in
-            guard tag.name == .emoji, !tag.otherParameters.isEmpty, let imageUrlString = tag.otherParameters.first, let imageUrl = URL(string: imageUrlString) else {
+            guard tag.name == .emoji, !tag.otherParameters.isEmpty, let imageURLString = tag.otherParameters.first, let imageURL = URL(string: imageURLString) else {
                 return nil
             }
 
-            return CustomEmoji(shortcode: tag.value, imageUrl: imageUrl)
+            return CustomEmoji(shortcode: tag.value, imageURL: imageURL)
         }
     }
 }
--- a/Sources/NostrSDK/CustomEmoji.swift
+++ b/Sources/NostrSDK/CustomEmoji.swift
@@ -19,6 +19,11 @@ public class CustomEmoji: CustomEmojiValidating, Equatable {
     /// A URL to the corresponding image file of the emoji.
     let imageURL: URL
 
+    /// ``Tag`` representation of the custom emoji.
+    var tag: Tag {
+        Tag(name: .emoji, value: shortcode, otherParameters: [imageURL.absoluteString])
+    }
+
     init?(shortcode: String, imageURL: URL) {
         self.shortcode = shortcode
         self.imageURL = imageURL
@@ -26,10 +31,6 @@ public class CustomEmoji: CustomEmojiValidating, Equatable {
         if !isValidShortcode(shortcode: shortcode) {
             return nil
         }
-    }
-
-    func tag() -> Tag {
-        Tag(name: .emoji, value: shortcode, otherParameters: [imageURL.absoluteString])
     }
 
     func isEqual(to customEmoji: CustomEmoji) -> Bool {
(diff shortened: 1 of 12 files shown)


Sources/NostrSDK/CustomEmoji.swift
```diff
@@ -0,0 +1,73 @@
...
+public class CustomEmoji: CustomEmojiValidating, Equatable {
+    public static func == (lhs: CustomEmoji, rhs: CustomEmoji) -> Bool {
+        lhs.isEqual(to: rhs)
+    }
+
+    /// A name given for the emoji, which MUST be comprised of only alphanumeric characters and underscores.
+    let shortcode: String
+
+    /// A a URL to the corresponding image file of the emoji.
+    let imageUrl: URL
```
tyiu, line 20: Is it `imageUrl` or `imageURL`?

Sources/NostrSDK/CustomEmoji.swift
```diff
@@ -0,0 +1,73 @@
...
+
+    init?(shortcode: String, imageUrl: URL) {
+        self.shortcode = shortcode
+        self.imageUrl = imageUrl
+
+        if !isValidShortcode(shortcode: shortcode) {
```
bryanmontz, line 26: Use `guard` and put this at the top of the initializer.

Sources/NostrSDK/CustomEmoji.swift
```diff
@@ -0,0 +1,73 @@
...
+            return nil
+        }
+    }
+
+    func tag() -> Tag {
```
bryanmontz, line 31: Can we use a `var` instead?
```swift
var tag: Tag {}
```

Sources/NostrSDK/CustomEmoji.swift
```diff
@@ -0,0 +1,73 @@
...
+
+            return CustomEmoji(shortcode: tag.value, imageUrl: imageUrl)
+        }
+    }
+}
+
+public protocol CustomEmojiValidating {}
+public extension CustomEmojiValidating {
+    /// Returns true if a custom emoji shortcode is valid (comprised of only alphanumeric characters and underscores)..
+    func isValidShortcode(shortcode: String) -> Bool {
```
bryanmontz, line 59: When the last word before the opening parenthesis is the same as the first argument, use an underscore for the argument label. That makes the call sites nicer.

```swift
func isValidShortcode(_ shortcode: String) -> Bool
```

Sources/NostrSDK/CustomEmoji.swift
```diff
@@ -0,0 +1,73 @@
...
+        let regex: NSRegularExpression
+        do {
+            regex = try NSRegularExpression(pattern: "/\\A[_0-9a-zA-Z]+\\z/")
+        } catch {
+            return false
+        }
+
+        guard regex.matches(in: shortcode, range: NSRange(location: 0, length: shortcode.count)).isEmpty else {
```
bryanmontz, line 67: Here you can simply return this expression:
```swift
return regex.matches(in: shortcode, range: NSRange(location: 0, length: shortcode.count)).isEmpty
```

Sources/NostrSDK/EventCreating.swift
```diff
@@ -173,7 +174,28 @@ public extension EventCreating {
...
+
+    /// Creates a ``ReactionEvent`` (kind 7) in response to a different ``NostrEvent`` and signs it with the provided ``Keypair``.
+    /// - Parameters:
+    ///   - customEmoji: The custom emoji to emojify with if the matching shortcode is found in the content field.
+    ///   - reactedEvent: The NostrEvent being reacted to.
+    ///   - keypair: The Keypair to sign with.
+    /// - Returns: The signed ``ReactionEvent``.
+    ///
+    /// See [NIP-25 - Reactions](https://github.com/nostr-protocol/nips/blob/master/25.md)
+    func reaction(withCustomEmoji customEmoji: CustomEmoji, reactedEvent: NostrEvent, signedBy keypair: Keypair) throws -> ReactionEvent {
```
bryanmontz, line 186: Do we need a separate function for this? Or could we add it to the existing one and make both the `content` and `customEmoji` optional with a default value of nil? I don't feel strongly either way.

Sources/NostrSDK/CustomEmoji.swift
bryanmontz: I don't really care too much, but my personal way is to do `imageURL` because "URL" is not a word, it's an acronym. When the name of a thing is just "url", I'll do `url`. Doesn't matter to me though.

bryanmontz (commented):
Great work, @tyiu!

Sources/NostrSDK/EventCreating.swift
tyiu: I think it's better to have a separate function. It's an XOR situation -- I want to avoid having the ability to define `content` that doesn't use the custom emoji shortcode from the emoji tag because that's out of spec.

Sources/NostrSDK/CustomEmoji.swift
tyiu: I'm getting this error for some reason. Is there any way I can solve it without needing to rename my variables?

```
Constant 'self.shortcode' used before being initialized
```

Sources/NostrSDK/CustomEmoji.swift
```diff
@@ -0,0 +1,73 @@
...
+/// [NIP-30 Custom Emoji](https://github.com/nostr-protocol/nips/blob/master/30.md)
+public class CustomEmoji: CustomEmojiValidating, Equatable {
+    public static func == (lhs: CustomEmoji, rhs: CustomEmoji) -> Bool {
+        lhs.isEqual(to: rhs)
+    }
+
+    /// A name given for the emoji, which MUST be comprised of only alphanumeric characters and underscores.
+    let shortcode: String
+
+    /// A a URL to the corresponding image file of the emoji.
```
tyiu, line 19: Typo: `A a`

Sources/NostrSDK/CustomEmoji.swift
tyiu: Great catch. I also realized that my logic is backwards and that this function isn't behaving as expected even when I reverse the logic. I'm missing a unit test for asserting behavior when the shortcode is invalid.

Sources/NostrSDK/CustomEmoji.swift
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+public protocol CustomEmojiValidating {}
+public extension CustomEmojiValidating {
+    /// Returns true if a custom emoji shortcode is valid (comprised of only alphanumeric characters and underscores)..
+    func isValidShortcode(shortcode: String) -> Bool {
+        let regex: NSRegularExpression
+        do {
+            regex = try NSRegularExpression(pattern: "/\\A[_0-9a-zA-Z]+\\z/")
```
tyiu, line 62: Pattern is wrong -- it doesn't need the leading and trailing `/`.

Sources/NostrSDK/CustomEmoji.swift
tyiu: Seems like I could just keep the guard at the bottom after the constants get initialized, though not ideal.

bryanmontz: Ah. It's because the function `isValidShortcode` requires that `self` exists, so it wants the required properties to be initialized when the function is called. Kinda weird, but I think it's worth it to keep the function separated from the type in that way.

Sources/NostrSDK/EventCreating.swift
bryanmontz: Yeah, it matches well if it's an XOR thing. I'm not familiar enough with these custom emojis.

bryanmontz (approved):

PR #108, opened 2023-11-18
Add support for NIP-52 date-based and time-based calendar events

tyiu:
This pull request does not implement everything from [NIP-52](https://github.com/nostr-protocol/nips/blob/master/52.md). There is enough complexity in this change that it warrants splitting the work up.

This PR focuses on only date-based calendar events and time-based calendar events. It does not implement calendars or RSVPs.

I tried to use protocols to break up tag parameters that may be shared across kinds and NIPs, so that they are composable and consistent in other places. I would love feedback on if this approach makes sense.

https://github.com/nostr-sdk/nostr-sdk-ios/issues/92


.swiftlint.yml
```diff
@@ -1,8 +1,10 @@
 disabled_rules:
-  - line_length
+  - cyclomatic_complexity
```
tyiu, line 2: It was complaining about the calendar event creation functions in `EventCreating` having more than 10 parameters.

.swiftlint.yml
```diff
@@ -1,8 +1,10 @@
...
   - identifier_name
-  - trailing_whitespace
+  - line_length
   - nesting
+  - trailing_whitespace
+  - type_body_length
```
tyiu, line 7: It was complaining that `EventCreatingTests` and `EventDecodingTests` exceeded 250 lines. Not sure if we care that much about it, but it's possible that we could split the file up.

Sources/NostrSDK/Tag.swift
```diff
@@ -156,6 +162,8 @@ public class Tag: Codable, Equatable {
             name = .subject
         case "emoji":
             name = .emoji
+        case "t":
```
tyiu, line 165: We should actually get rid of the concept of mapping tag names to enums as that model does not actually match reality. For example, the `r` tag can be reference or relay. I think we should just expose it as a raw string. I didn't want to make that change in this PR as it adds too much complexity and falls outside of the scope of what I was trying to implement.

.swiftlint.yml
bryanmontz: These are fine to opt out.

Sources/NostrSDK/Tag.swift
bryanmontz: I think I agree. This is a big bummer though and we should maybe consider a couple alternatives. You were right to not add that to this PR tho.

Sources/NostrSDK/Events/Calendars/CalendarEventParticipant.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    public init(pubkey: PublicKey, relay: URL? = nil, role: String? = nil) {
+        var otherParameters: [String] = [relay?.absoluteString ?? ""]
+        if let role, !role.isEmpty {
+            otherParameters.append(role)
+        }
+
+        tag = Tag(name: .pubkey, value: pubkey.hex, otherParameters: otherParameters)
+    }
+}
```
bryanmontz, line 67:
```suggestion
    public init(pubkey: PublicKey, relayURL: URL? = nil, role: String? = nil) {
        var otherParameters: [String] = [relayURL?.absoluteString ?? ""]
        if let role, !role.isEmpty {
            otherParameters.append(role)
        }

        tag = Tag(name: .pubkey, value: pubkey.hex, otherParameters: otherParameters)
    }
}
```

Sources/NostrSDK/Events/Calendars/DateBasedCalendarEventNostrEvent.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    /// Exclusive end date.
+    /// End date represented by ``DateComponents`` in the calendar context of ``Calendar.Identifier.iso8601``, with `year`, `month`, and `day` populated.
+    /// `nil` is returned if the backing `end` tag is malformed or if the calendar event ends on the same date as start.
+    public var end: DateComponents? {
+        guard let startString = tags.first(where: { $0.name.rawValue == "end" })?.value else {
+            return nil
+        }
+
+        return startString.dateStringAsDateComponents
```
bryanmontz, line 56:
```suggestion
        guard let endString = tags.first(where: { $0.name.rawValue == "end" })?.value else {
            return nil
        }

        return endString.dateStringAsDateComponents
```

Sources/NostrSDK/Events/Calendars/TimeBasedCalendarEventNostrEvent.swift
```diff
@@ -0,0 +1,85 @@
...
+    override init(kind: EventKind, content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+
+    public init(content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: .timeBasedCalendarEvent, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    /// Universally unique identifier (UUID).
+    public var uuid: String? {
```
bryanmontz, line 26: For the functions and computed vars that are the same in both types...

Maybe could have a parent class called `CalendarEvent` that has the shared code? Just a suggestion, because sometimes this introduces more trouble. Or maybe it could be implemented as a protocol extension to avoid class hierarchy.

Sources/NostrSDK/Events/Tags/PubkeyTag.swift
```diff
@@ -0,0 +1,14 @@
...
+//  PubkeyTag.swift
+//  
+//
+//  Created by Terry Yiu on 11/15/23.
+//
+
+import Foundation
+
+/// A public key tag.
+public protocol PubkeyTag {
```
bryanmontz, line 11: Since this protocol only requires that conformers provide a pubkey and not that they are a `Tag`, maybe this could be `PubkeyProviding`?

Sources/NostrSDK/Events/Tags/ReferenceTagInterpreting.swift
```diff
@@ -0,0 +1,18 @@
...
+//  ReferenceTagInterpreting.swift
+//
+//
+//  Created by Terry Yiu on 11/16/23.
+//
+
+import Foundation
+
+/// Interprets reference tags on events. References are URLs.
+public protocol ReferenceTagInterpreting: NostrEvent {}
```
bryanmontz, line 11: Yeah, I like this pattern.

Sources/NostrSDK/Events/Tags/RelayTagParameter.swift
```diff
@@ -0,0 +1,13 @@
...
+//  RelayTagParameter.swift
+//
+//
+//  Created by Terry Yiu on 11/15/23.
+//
+
+import Foundation
+
+/// A relay tag parameter. This parameter may appear as metadata to various tags, such as pubkey tags and event tags.
+public protocol RelayTagParameter {
```
bryanmontz, line 11: Just a suggestion, but maybe `RelayProviding`?

Sources/NostrSDK/Events/Tags/RelayTagParameter.swift
```diff
@@ -0,0 +1,13 @@
...
+    var relay: URL? { get }
```
bryanmontz, line 12: `relayURL`?

Sources/NostrSDK/Helpers/String+Additions.swift
```diff
@@ -31,4 +31,47 @@ public extension String {
 
         return data
     }
+
+    /// Returns ``DateComponents`` from a string representation of a date in the format of yyyy-mm-dd
+    /// or nil if the string does not match the format.
+    var dateStringAsDateComponents: DateComponents? {
```
bryanmontz, line 37: I would suggest that this instead be a failable initializer in an extension on `DateComponents`.

Would start like this:
```swift
extension DateComponents {
    convenience init?(dateString: String) { ... }
}
```

You can return nil inside that. Then you'd `guard let` or `if let` when you use the initializer:
```swift
if let comps = DateComponents(dateString: <val>) { ... }
```

Tests/NostrSDKTests/EventCreatingTests.swift
```diff
@@ -285,4 +285,107 @@ final class EventCreatingTests: XCTestCase, EventCreating, EventVerifying, Fixtu
...
+        XCTAssertEqual(timeBasedCalendarEvent.startTimeZone, startTimeZone)
+        XCTAssertEqual(timeBasedCalendarEvent.endTimeZone, endTimeZone)
+        XCTAssertEqual(timeBasedCalendarEvent.location, location)
+        XCTAssertEqual(timeBasedCalendarEvent.geohash, geohash)
+        XCTAssertEqual(timeBasedCalendarEvent.participants, participants)
+        XCTAssertEqual(timeBasedCalendarEvent.hashtags, hashtags)
+        XCTAssertEqual(timeBasedCalendarEvent.references, references)
+
+        try verifyEvent(timeBasedCalendarEvent)
+    }
```
bryanmontz, line 390: It would be nice to add a test to confirm that the function fails when end is before start.

bryanmontz (commented):
This is some really great work, @tyiu!

Sources/NostrSDK/EventCreating.swift
```diff
@@ -225,4 +225,150 @@ public extension EventCreating {
...
+        }
+
+        let endDate = end?.date
+        // If the end date is ommitted, the calendar event ends on the same date as the start date.
+        if endDate != nil {
+            // The start date must occur before the end date, if it exists.
+            guard let endDate, startDate < endDate else {
+                throw EventCreatingError.invalidInput
+            }
+        }
```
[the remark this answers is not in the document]
tyiu: This would break the second `if let endDate {` statement further down. I was trying to perform validations before creating the tags array, and also I wanted to keep a certain order of the tags so that it would be easier to read for debugging purposes. I think what I'll do is I'll fold the validation in with the end date tag construction in the same let block, and then I'll arrange the ordering.

Sources/NostrSDK/EventKind.swift
```diff
@@ -67,7 +67,15 @@ public enum EventKind: RawRepresentable, CaseIterable, Codable, Equatable {
     ///
     /// See [NIP-56](https://github.com/nostr-protocol/nips/blob/b4cdc1a73d415c79c35655fa02f5e55cd1f2a60c/56.md#nip-56).
     case report
-    
+
+    /// This kind of event represents an occurrence that spans between a start time and end time.
```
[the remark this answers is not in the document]
tyiu: Yes! Nice catch.

Sources/NostrSDK/Events/Calendars/DateBasedCalendarEventNostrEvent.swift
```diff
@@ -0,0 +1,68 @@
...
+//
+//  Created by Terry Yiu on 11/13/23.
+//
+
+import Foundation
+
+/// Date-based calendar event starts on a date and ends before a different date in the future.
+/// Its use is appropriate for all-day or multi-day events where time and time zone hold no significance. e.g., anniversary, public holidays, vacation days.
+/// See [NIP-52](https://github.com/nostr-protocol/nips/blob/master/52.md).
+public final class DateBasedCalendarEventNostrEvent: NostrEvent, CalendarEventParticipantInterpreting, HashtagInterpreting, ReferenceTagInterpreting {
```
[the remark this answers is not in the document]
tyiu: That might be simpler.

I was trying to consider what the naming would be once I bring in kind 31924 "Calendars" (wrapper around date-based and time-based calendar events). If we call it `CalendarEvent`, that's overloading the term. Perhaps we go with your suggestion of dropping the `NostrEvent` suffix for `DateBasedCalendarEvent` and `TimeBasedCalendarEvent`, and then we'll call calendars as `CalendarNostrEvent`.

Sources/NostrSDK/Events/Calendars/DateBasedCalendarEventNostrEvent.swift
tyiu: Copy paste error. Thanks!

Sources/NostrSDK/Events/Calendars/TimeBasedCalendarEventNostrEvent.swift
tyiu: That's a great idea. It'll make it easier for clients to integrate without duplicating as much code.

Sources/NostrSDK/Helpers/String+Additions.swift
tyiu: Oh, that's a way better pattern.

Sources/NostrSDK/Tag.swift
tyiu: This isn't a well thought through idea, but maybe each event kind can define their own set of known named tags somehow.

Sources/NostrSDK/Events/Tags/RelayTagParameter.swift
tyiu: This was a great suggestion. I was able to change `RecommendServerEvent` to conform to the protocol as well.

Sources/NostrSDK/Helpers/String+Additions.swift
tyiu: Done! `convenience` doesn't apply to structs, only classes, so I left that out.

Sources/NostrSDK/EventCreating.swift
```diff
@@ -225,4 +225,151 @@ public extension EventCreating {
...
+
+            tags.append(Tag(name: .unknown("end"), value: iso8601DateFormatter.string(from: endDate)))
+        }
+
+        // Re-arrange tags so that it's easier to read with the identifier and name appearing first in the list of tags,
+        // and the end date being placed next to the start date.
+        tags = [
+            Tag(name: .unknown("d"), value: UUID().uuidString),
+            Tag(name: .unknown("name"), value: name),
+            Tag(name: .unknown("start"), value: iso8601DateFormatter.string(from: startDate))
```
tyiu, line 281: I discovered a bug where if `startDate` or `endDate` are in a non-UTC time zone, then what gets computed is actually shifted to be in UTC time. So the intended date in the input could be one day off from what gets added to the tag.

- [ ] Fix time zone issue with date components

tyiu:
I addressed all the feedback and known issues. This PR is ready for another review.

Sources/NostrSDK/Events/Calendars/TimeOmittedDate.swift
```diff
@@ -0,0 +1,92 @@
...
+
+        if lhs.month < rhs.month {
+            return true
+        }
+
+        if lhs.month > rhs.month {
+            return false
+        }
+
+        return lhs.day < rhs.day
```
bryanmontz, line 29: ChatGPT suggests this simpler alternative:
```swift
if lhs.year != rhs.year {
        return lhs.year < rhs.year
    }

    if lhs.month != rhs.month {
        return lhs.month < rhs.month
    }

    return lhs.day < rhs.day
```

Sources/NostrSDK/Events/Calendars/CalendarEventParticipant.swift
```diff
@@ -54,6 +56,11 @@ public struct CalendarEventParticipant: PubkeyProviding, RelayProviding, Equatab
         self.tag = pubkeyTag
     }
 
+    /// Initializes a calendar event participant.
+    /// - Parameters:
+    ///   - pubkey: The public key of the participant.
+    ///   - relayURL: A relay in which the participant can be found
```
bryanmontz, line 62: Missing period at the end. For some reason GitHub wouldn't let me do a suggestion here.

Sources/NostrSDK/Events/Calendars/TimeOmittedDate.swift
tyiu: Much nicer. Fixed!

Sources/NostrSDK/Events/Calendars/CalendarEventParticipant.swift
tyiu: Fixed!

bryanmontz (approved):

PR #110, opened 2023-11-26
NIP-23: add support for Longform Content Events

bryanmontz:
Closes https://github.com/nostr-sdk/nostr-sdk-ios/issues/93

@tyiu I elected to avoid the issue of changing how tags work (currently an enum) because I had already started this work. After this PR is merged, I will work on that.


Sources/NostrSDK/EventCreating.swift
```diff
@@ -225,6 +225,52 @@ public extension EventCreating {
         ]
         return try ReportEvent(content: additionalInformation, tags: tags, signedBy: keypair)
     }
+    
+    /// Creates a ``LongformContentEvent`` (kind 30023, a parameterized replaceable event) for long-form text content, generally refered to as "articles" or "blog posts".
```
tyiu, line 229:
```suggestion
    /// Creates a ``LongformContentEvent`` (kind 30023, a parameterized replaceable event) for long-form text content, generally referred to as "articles" or "blog posts".
```

Sources/NostrSDK/Events/LongformContentEvent.swift
```diff
@@ -0,0 +1,54 @@
+//
+//  LongformContentEvent.swift
+//
+//
+//  Created by Bryan Montz on 11/2/23.
+//
+
+import Foundation
+
+public final class LongformContentEvent: NostrEvent {
```
tyiu, line 10: We should probably add documentation to the class and its attributes as well.

Sources/NostrSDK/Tag.swift
```diff
@@ -8,51 +8,82 @@
...
     /// points to the id of an event this event is quoting, replying to or referring to somehow
     case event
     
+    /// a hashtag to categorize events for easy searching
+    case hashtag
+    
     /// points to a pubkey of someone that is referred to in the event
     case pubkey
     
+    case publishedAt
```
tyiu, line 25: As mentioned in your PR description, adding tags here for now should be fine. But I foresee a better solution with moving to kind-specific tags in the specific Nostr event class file, rather than having a universal catch-all.

tyiu (commented):
I like that we're not being prescriptive about providing a markdown abstraction, but rather just exposing the raw string and letting the developer choose how they want to approach it.

Overall looks great! Just a few minor nitpicks.

Sources/NostrSDK/EventCreating.swift
bryanmontz: Thank you!

Sources/NostrSDK/Events/LongformContentEvent.swift
bryanmontz: Added. Thanks.

tyiu (approved):

PR #113, opened 2023-11-29
Fix tags to not have a mapping from name and TagName enum

tyiu:
There isn't a one-to-one mapping. For example, the "r" tag can have multiple meanings and semantics. So it makes sense to be able to map a tag name enum to the actual string representation, but not in the other direction.

### DIFF
--- a/Sources/NostrSDK/Events/Calendars/DateBasedCalendarEvent.swift
+++ b/Sources/NostrSDK/Events/Calendars/DateBasedCalendarEvent.swift
@@ -28,7 +28,7 @@ public final class DateBasedCalendarEvent: NostrEvent, CalendarEventInterpreting
     /// Start date is represented by ``TimeOmittedDate``.
     /// `nil` is returned if the backing `start` tag is malformed.
     public var startDate: TimeOmittedDate? {
-        guard let startString = tags.first(where: { $0.name.rawValue == "start" })?.value else {
+        guard let startString = tags.first(where: { $0.name == "start" })?.value else {
             return nil
         }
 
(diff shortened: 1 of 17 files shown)


Sources/NostrSDK/Events/Calendars/DateBasedCalendarEvent.swift
```diff
@@ -28,7 +28,7 @@ public final class DateBasedCalendarEvent: NostrEvent, CalendarEventInterpreting
     /// Start date is represented by ``TimeOmittedDate``.
     /// `nil` is returned if the backing `start` tag is malformed.
     public var startDate: TimeOmittedDate? {
-        guard let startString = tags.first(where: { $0.name.rawValue == "start" })?.value else {
+        guard let startString = tags.first(where: { $0.name == "start" })?.value else {
```
tyiu, line 31: It might be outside of the scope of this PR, but I wonder if it would be useful to have bespoke tag name enums scoped down to each event kind level. That way we're not dealing with raw strings everywhere, and it can be shared between `EventCreating` and the NostrEvent kind class.

tyiu, line 31: It might be outside of the scope of this PR, but I wonder if it would be useful to have bespoke tag name enums scoped down to each event kind level. That way we're not dealing with raw strings everywhere, and it can be shared between `EventCreating` and the NostrEvent kind class.

bryanmontz (approved):
Looks good. I'm still thinking about how this could be improved a bit further, maybe something along the lines of what you said about having per-event kind tags.

PR #119, opened 2023-12-16
Refactor and add tag interpreters for identifiers, titles, relays, and replaceable events

tyiu:


Sources/NostrSDK/Tag.swift
```diff
@@ -15,7 +15,10 @@ public enum TagName: String {
     
     /// points to the id of an event this event is quoting, replying to or referring to somehow
     case event = "e"
-    
+
+    /// coordinates to a replaceable event, which includes the kind number, pubkey that signed the event, and optionally the identifier (if the replaceable event is parameterized)
+    case replaceableEvent = "a"
```
bryanmontz, line 20: Maybe we could call this `eventCoordinates`? The NIPs repo's README says "coordinates to an event" as the value for "a" tags.

Sources/NostrSDK/Events/Tags/ReplaceableEventCoordinates.swift
```diff
@@ -0,0 +1,118 @@
...
+/// See [NIP-01 Tags](https://github.com/nostr-protocol/nips/blob/master/01.md#tags).
+public struct ReplaceableEventCoordinates: PubkeyProviding, RelayProviding, Equatable {
+    public static func == (lhs: Self, rhs: Self) -> Bool {
+        lhs.tag == rhs.tag
+    }
+
+    /// The tag representation of these replaceable event coordinates.
+    public let tag: Tag
+
+    private var tagValueSplit: [Substring] {
```
bryanmontz, line 20: Naming suggestion: "tagComponents".

Sources/NostrSDK/Events/Tags/ReplaceableEventCoordinates.swift
```diff
@@ -0,0 +1,118 @@
...
+        tag.value.split(separator: ":", omittingEmptySubsequences: false)
+    }
+
+    /// The kind integer of the referenced replaceable event.
+    /// Returns `nil` if the kind integer part of the tag value is malformed.
+    public var kind: EventKind? {
+        guard let kindInt = Int(String(tagValueSplit[0])) else {
```
bryanmontz, line 27: All direct array access like this should be protected against crashes like this:

If accessing `index`:
```swift
guard index < tagValueSplit.count else ...
```

Sources/NostrSDK/Events/Tags/ReplaceableEventCoordinates.swift
```diff
@@ -0,0 +1,118 @@
...
+            return nil
+        }
+
+        return EventKind(rawValue: kindInt)
+    }
+
+    /// The pubkey that signed the referenced replaceable event.
+    /// Returns `nil` if the pubkey part of the tag value is malformed.
+    public var pubkey: PublicKey? {
+        PublicKey(hex: String(tagValueSplit[1]))
```
bryanmontz, line 37: Add direct array access protection.

Sources/NostrSDK/Events/Tags/ReplaceableEventCoordinates.swift
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /// The identifier of the referenced replaceable event.
+    /// Returns `nil` if the returned event is not a parameterized replaceable event.
+    public var identifier: String? {
+        let identifierParameter = tagValueSplit[2]
```
bryanmontz, line 43: Same here.

Sources/NostrSDK/Events/Tags/ReplaceableEventCoordinates.swift
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        return relayString.relayURL
+    }
+
+    /// Initializes coordinates to a replaceable event from a ``Tag``.
+    /// For a parameterized replaceable event, a tag value of `<kind integer>:<32-bytes lowercase hex of a pubkey>:<d tag value>` is expected.
+    /// For a non-parameterized replaceable event, a tag value of `<kind integer>:<32-bytes lowercase hex of a pubkey>:` is expected.
+    ///
+    /// Returns `nil` if the tag is not a replaceable event tag  or if the tag value does not have at least two ":" colon separators.
```
bryanmontz, line 65:
```suggestion
    /// Returns `nil` if the tag is not a replaceable event tag or if the tag value does not have at least two ":" colon separators.
```

Sources/NostrSDK/Events/Tags/ReplaceableEventCoordinates.swift
```diff
@@ -0,0 +1,118 @@
...
+    ///
+    /// - Parameters:
+    ///   - pubkey: The public key of the participant.
```
bryanmontz, line 68: The word "participant" pertains specifically to calendar events, right? Maybe could be more generic here.

Sources/NostrSDK/Events/Tags/ReplaceableEventCoordinates.swift
```diff
@@ -0,0 +1,118 @@
...
+        guard split.count >= 3 else {
+            return nil
+        }
+
+        self.tag = replaceableEventTag
+    }
+
+    /// Initializes coordinates to a replaceable event.
+    /// - Parameters:
+    ///   - kind: The kind integer of the referenced replaceable event.
```
bryanmontz, line 85:
```suggestion
    ///   - kind: The ``EventKind`` of the referenced replaceable event.
```

bryanmontz (commented):
Thanks, @tyiu!

Sources/NostrSDK/Events/Tags/ReplaceableEventCoordinates.swift
tyiu: Oops, that's a copy paste error. Nice catch.

Sources/NostrSDK/Events/Tags/TitleTagInterpreting.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  TitleTagInterpreting.swift
+//  
+//
+//  Created by Terry Yiu on 12/16/23.
+//
+
+import Foundation
+
+public protocol TitleTagInterpreting: NostrEvent {}
```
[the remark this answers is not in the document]
tyiu: My understanding is that `title` is a common tag. I agree that it's starting to feel a bit bloated to have one for each tag.

My concern with something like `CommonTagInterpreting` is that, although the tags all might be commonly used, not all kinds use all common tags, and some may use just a subset.

Need to think about this more, but for the time being, this is probably ok for now.

Happy to reconsider if you feel strongly about it.

Sources/NostrSDK/Helpers/String+Additions.swift
```diff
@@ -8,7 +8,19 @@
 import Foundation
 
 public extension String {
-    
+
+    /// Transforms a string into a Nostr relay ``URL`` if the URL format appears valid.
+    var relayURL: URL? {
```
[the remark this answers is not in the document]
tyiu: Yeah, I felt a bit gross doing this. I think the direction that your suggestion moves to is much better. I'll rework it.

Sources/NostrSDK/Tag.swift
tyiu: That's actually what I had initially and changed to `replaceableEvent`. Not sure if I like the naming in the NIPs repo, but going back to our core principles, we should just follow the naming from the spec. I'll change it back to `eventCoordinates`.

Sources/NostrSDK/Events/Tags/ReplaceableEventCoordinates.swift
tyiu: You're right. I thought that the initializer guard would be sufficient, but it's better to be safe to check here as well.

Sources/NostrSDK/Events/Tags/TitleTagInterpreting.swift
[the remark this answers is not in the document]
bryanmontz: Naw, good for now.

Sources/NostrSDK/Helpers/String+Additions.swift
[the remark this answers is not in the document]
tyiu: Oh, one issue with this approach is that I can't use the validate call in initializers.

```
'self' used before 'self.init' call or assignment to 'self'
```

Do you know what the best practice is here?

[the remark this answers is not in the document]
bryanmontz: Ugh, yeah that's an annoying limitation.

Unfortunately the only way I know to workaround it is like this:
```swift
final class RelayURLValidator: RelayURLValidating {
    static let shared = RelayURLValidator()
}
```

Then at your call site in the init:
```swift
guard let relayURL = try? RelayURLValidator.shared.validateRelayURLString(testString) else {
    return
}
// use valid relayURL
```

It's not great and kind of defeats the purpose of the protocol extension. But that's generally what I do.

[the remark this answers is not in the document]
tyiu: Fixed. Please take another look.

bryanmontz (approved):

PR #120, opened 2023-12-17
add support for NIP-51 mute lists

bryanmontz:
Implements the first NIP-51 list event: a kind 10000 MuteListEvent.
https://github.com/nostr-protocol/nips/blob/master/51.md#standard-lists

Other list types will follow in a subsequent PR.

### DIFF
--- a/Sources/NostrSDK/EventCreating.swift
+++ b/Sources/NostrSDK/EventCreating.swift
@@ -229,23 +229,23 @@ public extension EventCreating {
     /// Creates a ``MuteListEvent`` (kind 10000) containing things the user doesn't want to see in their feeds. Mute list items be publicly visible or private.
     /// - Parameters:
     ///   - publiclyMutedPubkeys: Pubkeys to mute.
-    ///   - secretlyMutedPubkeys: Pubkeys to secretly mute.
+    ///   - privatelyMutedPubkeys: Pubkeys to secretly mute.
     ///   - publiclyMutedEventIds: Event ids to mute.
-    ///   - secretlyMutedEventIds: Event ids to secretly mute.
+    ///   - privatelyMutedEventIds: Event ids to secretly mute.
     ///   - publiclyMutedHashtags: Hashtags to mute.
-    ///   - secretlyMutedHashtags: Hashtags to secretly mute.
+    ///   - privatelyMutedHashtags: Hashtags to secretly mute.
     ///   - publiclyMutedKeywords: Keywords to mute.
-    ///   - secretlyMutedKeywords: Keywords to secretly mute.
+    ///   - privatelyMutedKeywords: Keywords to secretly mute.
     ///   - keypair: The Keypair to sign with.
     /// - Returns: The signed ``MuteListEvent``.
     func muteList(withPubliclyMutedPubkeys publiclyMutedPubkeys: [String] = [],
-                  secretlyMutedPubkeys: [String] = [],
+                  privatelyMutedPubkeys: [String] = [],
                   publiclyMutedEventIds: [String] = [],
-                  secretlyMutedEventIds: [String] = [],
+                  privatelyMutedEventIds: [String] = [],
                   publiclyMutedHashtags: [String] = [],
-                  secretlyMutedHashtags: [String] = [],
+                  privatelyMutedHashtags: [String] = [],
                   publiclyMutedKeywords: [String] = [],
-                  secretlyMutedKeywords: [String] = [],
+                  privatelyMutedKeywords: [String] = [],
                   signedBy keypair: Keypair) throws -> MuteListEvent {
         var publicTags = [Tag]()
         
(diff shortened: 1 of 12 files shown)


Sources/NostrSDK/Events/MuteListEvent.swift
```diff
@@ -0,0 +1,83 @@
...
+        allValues(forTagName: .pubkey) ?? []
+    }
+    
+    /// The publicly muted event ids (threads).
+    public var eventIds: [String] {
+        allValues(forTagName: .event) ?? []
+    }
+    
+    /// The publicly muted hashtags.
+    public var hashtags: [String] {
```
tyiu, line 38: I'm surprised that the compiler didn't complain about `hashtags` being declared twice due to this class extending `HashtagInterpreting`.

On a related note, we can update `HashtagInterpreting` and a few other classes to use the `allValues` helper. If you didn't want to add noise and bloat to this PR, we can do that in a separate PR as a small refactor.

Sources/NostrSDK/Events/MuteListEvent.swift
```diff
@@ -0,0 +1,83 @@
...
+    override init(kind: EventKind, content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    init(content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: .muteList, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    /// The publicly muted public keys (authors).
+    public var pubkeys: [String] {
```
tyiu, line 28: Wondering if it makes sense to have something like a `PubkeyTagInterpreting` as there's a few Nostr events that do it.

However, one issue that I see with refactoring in this way is that we lose some of the finer-grained documentation on the kind level.

Sources/NostrSDK/Events/MuteListEvent.swift
```diff
@@ -0,0 +1,83 @@
...
+        allValues(forTagName: .pubkey) ?? []
+    }
+    
+    /// The publicly muted event ids (threads).
+    public var eventIds: [String] {
```
tyiu, line 33: Same thought here.

tyiu (commented):
Looks great overall! Left some comments.

Sources/NostrSDK/Events/NostrEvent.swift
```diff
@@ -99,4 +99,11 @@ public class NostrEvent: Codable {
     public func valueForRawTagName(_ tagName: String) -> String? {
```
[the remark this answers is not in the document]
bryanmontz: Agree. I changed the documentation, but I think this warrants the function names including "first". Thanks.

Sources/NostrSDK/Events/NostrEvent.swift
```diff
@@ -99,4 +99,11 @@ public class NostrEvent: Codable {
...
         tags.first(where: { $0.name == tagName })?.value
     }
+    
+    /// All tags with the provided name.
+    /// - Parameter tag: The tag name to filter.
+    /// - Returns: The values associated with the tags of the provided name.
+    public func allValues(forTagName tag: TagName) -> [String]? {
```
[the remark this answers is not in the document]
bryanmontz: Yeah, in a separate PR I'll update all occurrences where this is applicable.

Sources/NostrSDK/Events/MuteListEvent.swift
bryanmontz: In a protocol extension, the function in the extension is just the "default" implementation and can be overridden by conformers.

Yep, coming in a separate PR.

bryanmontz: I'll remove this function's implementation though, since it's just duplicating the one in the protocol.

Sources/NostrSDK/Events/MuteListEvent.swift
bryanmontz: Agree. It would be funny if you could override *just* the documentation. I think it's a good idea to move to a protocol for interpreting, but I'll do this in a next PR for nothing but tags improvements.

Sources/NostrSDK/Events/MuteListEvent.swift
bryanmontz: Yep, same.

Sources/NostrSDK/EventCreating.swift
```diff
@@ -226,6 +226,69 @@ public extension EventCreating {
         return try ReportEvent(content: additionalInformation, tags: tags, signedBy: keypair)
     }
     
+    /// Creates a ``MuteListEvent`` (kind 10000) containing things the user doesn't want to see in their feeds. Mute list items be publicly visible or private.
+    /// - Parameters:
+    ///   - publiclyMutedPubkeys: Pubkeys to mute.
+    ///   - secretlyMutedPubkeys: Pubkeys to secretly mute.
```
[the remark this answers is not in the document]
bryanmontz: Yeah I think "private" is probably better. I was thinking about it from the client application's call site, but better to follow the spec we're implementing here. Will change.

bryanmontz:
@tyiu Feedback addressed!

tyiu (approved):

PR #122, opened 2023-12-19
Tags: add shorthand functions for easier usage and readability

bryanmontz:
We can now optionally use shorthand functions for creating tags.

```swift
let tag = Tag.pubkey(<pubkey hex>)

let tags: [Tag] = [.pubkey(<pubkey hex>)]
```

I created these functions for tags that are used frequently, but more could be added. The regular initializer is unchanged and can still be used when desired.


tyiu (approved):
Much cleaner. Thanks!

PR #124, opened 2023-12-22
Update README

tyiu:


bryanmontz (approved):
This is good, but the class name should be changed as well.

tyiu:
> This is good, but the class name should be changed as well.

Yep, I'll do that in a separate PR.

tyiu:
> > This is good, but the class name should be changed as well.
> 
> Yep, I'll do that in a separate PR.

https://github.com/nostr-sdk/nostr-sdk-ios/pull/125

PR #126, opened 2023-12-23
Add deletion request support for replaceable events

tyiu:
I also restructured the code to make a clearer distinction between parameterized and non-parameterized replaceable events.

Relevant NIPs:
https://github.com/nostr-protocol/nips/blob/master/01.md
https://github.com/nostr-protocol/nips/blob/master/09.md


Sources/NostrSDK/Events/Tags/IdentifierTagInterpreting.swift
```diff
@@ -1,29 +0,0 @@
-//
-//  IdentifierTagInterpreting.swift
-//  
-//
-//  Created by Terry Yiu on 12/16/23.
-//
-
-import Foundation
-
-public protocol IdentifierTagInterpreting: NostrEvent {}
```
tyiu, line 10: Renamed to `ReplaceableEvent`.

Sources/NostrSDK/Events/ReplaceableEvent.swift
```diff
@@ -0,0 +1,22 @@
...
+//  ReplaceableEvent.swift
+//
+//
+//  Created by Terry Yiu on 12/23/23.
+//
+
+import Foundation
+
+/// Replaceable events are ones where only the latest event MUST be stored by relays, older versions MAY be discarded.
+public protocol ReplaceableEvent: NostrEvent {
```
tyiu, line 11: Renamed from `IdentifierTagInterpreting`.

Sources/NostrSDK/Events/Tags/IdentifierTagInterpreting.swift
```diff
@@ -1,29 +0,0 @@
...
-    /// This identifier is represented by the "d" tag, which is distinctly different from the `id` field on ``NostrEvent``.
-    var identifier: String? {
-        firstValueForTagName(.identifier)
-    }
-
-    /// The event coordinates that can be used to fetch this replaceable event's kind, pubkey, and identifier from a relay.
-    ///
-    /// - Parameters:
-    ///   - relayURL: A relay URL that this replaceable event could be found.
-    func identifierEventCoordinates(_ relayURL: URL? = nil) -> EventCoordinates? {
```
tyiu, line 22: Non-parameterized replaceable events do not have identifiers but they do have event coordinates, so the way this protocol was structured didn't make sense.

Sources/NostrSDK/EventKind.swift
```diff
@@ -145,4 +145,25 @@ public enum EventKind: RawRepresentable, CaseIterable, Codable, Equatable {
...
     }
+
+    /// For kind `n` such that `10000 <= n < 20000 || n == 0 || n == 3`, events are replaceable,
+    /// which means that, for each combination of pubkey and kind,
+    /// only the latest event MUST be stored by relays, older versions MAY be discarded.
+    ///
+    /// See [NIP-01 - Kinds](https://github.com/nostr-protocol/nips/blob/master/01.md#kinds).
+    public var isNonParameterizedReplaceable: Bool {
+        switch rawValue {
+        case 10000..<20000, 0, 3: return true
```
bryanmontz, line 156: Excellent use of `switch-case`!

Sources/NostrSDK/Events/Tags/EventCoordinates.swift
```diff
@@ -85,13 +85,49 @@ public struct EventCoordinates: PubkeyProviding, RelayProviding, RelayURLValidat
...
     /// - Parameters:
-    ///   - kind: The ``EventKind`` of the referenced replaceable event.
-    ///   - pubkey: The pubkey that signed the referenced replaceable event.
-    ///   - identifier: The identifier of the referenced replaceable event.
-    ///   - relayURL: A relay in which the referenced replaceable event could be found.
-    public init?(kind: EventKind, pubkey: PublicKey, identifier: String?, relayURL: URL? = nil) {
+    ///   - kind: The ``EventKind`` of the referenced non-parameterized replaceable event.
+    ///   - pubkey: The pubkey that signed the referenced non-parameterized replaceable event.
+    ///   - relayURL: A relay in which the referenced non-parameterized replaceable event could be found.
+    public init?(kind: EventKind, pubkey: PublicKey, relayURL: URL? = nil) {
```
bryanmontz, line 94: Couldn't we have `identifier` still here and default it to empty string if not provided? Then this initializer could be used for either kind? You could still validate the "parameterized or not" part based on the kind. I might be missing something here.

Also I think it would be confusing to only return nil here. IMO we should either put an assert here with a relevant message or have this initializer throw an error with a human-readable message.

Sources/NostrSDK/Events/NonParameterizedReplaceableEvent.swift
```diff
@@ -0,0 +1,27 @@
...
+public extension NonParameterizedReplaceableEvent {
+    var replaceableEventCoordinates: EventCoordinates? {
+        guard kind.isNonParameterizedReplaceable, let publicKey = PublicKey(hex: pubkey) else {
+            return nil
+        }
+
+        return EventCoordinates(kind: kind, pubkey: publicKey, relayURL: nil)
+    }
+
+    func replaceableEventCoordinates(relayURL: URL) -> EventCoordinates? {
```
bryanmontz, line 20: Just make `relayURL` optional with a default value of nil. Then you don't need the `var` above at all.

Sources/NostrSDK/Events/ParameterizedReplaceableEvent.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    var replaceableEventCoordinates: EventCoordinates? {
+        guard kind.isParameterizedReplaceable, let identifier, let publicKey = PublicKey(hex: pubkey) else {
+            return nil
+        }
+
+        return EventCoordinates(kind: kind, pubkey: publicKey, identifier: identifier, relayURL: nil)
+    }
+
+    func replaceableEventCoordinates(relayURL: URL) -> EventCoordinates? {
```
bryanmontz, line 26: Same comment here as for the non-parameterized one.

Sources/NostrSDK/Events/NonParameterizedReplaceableEvent.swift
tyiu: I wanted to do that but the issue is that you can't assign default values on the protocol level. In this case, it's the `ReplaceableEvent` protocol. So it makes for a cumbersome call of `event.replaceableEventCoordinates(relayURL: nil)`. Which might be fine, actually. We generally want to encourage specifying a relay for the "gossip model" rather than omitting it.

Sources/NostrSDK/Events/NonParameterizedReplaceableEvent.swift
```diff
@@ -0,0 +1,23 @@
...
+    func replaceableEventCoordinates(relayURL: URL? = nil) -> EventCoordinates? {
+        guard kind.isNonParameterizedReplaceable, let publicKey = PublicKey(hex: pubkey) else {
+            return nil
+        }
+
+        do {
+            return try EventCoordinates(kind: kind, pubkey: publicKey, relayURL: relayURL)
+        } catch {
+            return nil
+        }
```
bryanmontz, line 21: This is identical to:
```suggestion
        return try? EventCoordinates(kind: kind, pubkey: publicKey, relayURL: relayURL)
```

Essentially you're saying you don't care about the thrown error.

Sources/NostrSDK/Events/ParameterizedReplaceableEvent.swift
```diff
@@ -0,0 +1,29 @@
...
+    func replaceableEventCoordinates(relayURL: URL? = nil) -> EventCoordinates? {
+        guard kind.isParameterizedReplaceable, let identifier, let publicKey = PublicKey(hex: pubkey) else {
+            return nil
+        }
+
+        do {
+            return try EventCoordinates(kind: kind, pubkey: publicKey, identifier: identifier, relayURL: relayURL)
+        } catch {
+            return nil
+        }
```
bryanmontz, line 27: Same.
```suggestion
        return try? EventCoordinates(kind: kind, pubkey: publicKey, identifier: identifier, relayURL: relayURL)
```

Sources/NostrSDK/Events/Tags/EventCoordinates.swift
```diff
@@ -86,16 +90,21 @@ public struct EventCoordinates: PubkeyProviding, RelayProviding, RelayURLValidat
...
+        }
+
         let otherParameters: [String]
         if let relayURL {
             guard (try? RelayURLValidator.shared.validateRelayURL(relayURL)) != nil else {
-                return nil
+                throw EventCoordinatesError.invalidInput
             }
 
             otherParameters = [relayURL.absoluteString]
```
bryanmontz, line 110: It would be better to allow `validateRelayURL` to throw it's own, more specific errors than to throw them away and throw a more generic one here.

```swift
            let validatedURL = try RelayURLValidator.shared.validateRelayURL(relayURL)
            otherParameters = [validatedURL.absoluteString]
```

Tests/NostrSDKTests/EventCreatingTests.swift
```diff
@@ -638,13 +689,16 @@ final class EventCreatingTests: XCTestCase, EventCreating, EventVerifying, Fixtu
...
+        XCTAssertEqual(calendarEventRSVP.replaceableEventCoordinates(relayURL: nil), expectedReplaceableEventCoordinates)
+
         try verifyEvent(calendarEventRSVP)
     }
 
     func testCalendarEventRSVPWithInvalidCalendarEventCoordinatesShouldFail() throws {
         let identifier = "hockey-practice-rsvp"
         let note = "Don't forget your skates!"
-        let eventCoordinates = try XCTUnwrap(EventCoordinates(kind: EventKind.textNote, pubkey: Keypair.test.publicKey, identifier: "abc"))
+        let eventCoordinates = try XCTUnwrap(EventCoordinates(kind: EventKind.longformContent, pubkey: Keypair.test.publicKey, identifier: "abc"))
```
bryanmontz, line 701:
```suggestion
        let eventCoordinates = try XCTUnwrap(EventCoordinates(kind: .longformContent, pubkey: Keypair.test.publicKey, identifier: "abc"))
```

Tests/NostrSDKTests/EventCreatingTests.swift
```diff
@@ -598,15 +646,15 @@ final class EventCreatingTests: XCTestCase, EventCreating, EventVerifying, Fixtu
         let identifier = "family-calendar"
         let title = "Family Calendar"
         let description = "All family events."
-        let eventCoordinates = try XCTUnwrap(EventCoordinates(kind: EventKind.textNote, pubkey: Keypair.test.publicKey, identifier: "abc"))
+        let eventCoordinates = try XCTUnwrap(EventCoordinates(kind: EventKind.longformContent, pubkey: Keypair.test.publicKey, identifier: "abc"))
```
bryanmontz, line 649:
```suggestion
        let eventCoordinates = try XCTUnwrap(EventCoordinates(kind: .longformContent, pubkey: Keypair.test.publicKey, identifier: "abc"))
```

bryanmontz (approved):

PR #128, opened 2023-12-25
add support for kind 10003 bookmarks lists

bryanmontz:


Sources/NostrSDK/EventCreating.swift
```diff
@@ -291,6 +291,67 @@ public extension EventCreating {
...
+            publicTags.append(coordinates.tag)
+        }
+        for hashtag in publiclyBookmarkedHashtags {
+            publicTags.append(.hashtag(hashtag))
+        }
+        for link in publiclyBookmarkedLinks {
+            publicTags.append(Tag(name: .webURL, value: link.absoluteString))
+        }
+        
+        var secretTags = [[String]]()
```
tyiu, line 328: It makes me a bit nervous to handle raw strings (as opposed to how we're handling the public tags) because we lose out on the existing structure and validation.

Sources/NostrSDK/Events/BookmarksListEvent.swift
```diff
@@ -0,0 +1,71 @@
...
+        try super.init(kind: .bookmarksList, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    /// Ids of bookmarked kind-1 notes.
+    public var noteIds: [String] {
+        allValues(forTagName: .event) ?? []
+    }
+    
+    /// Coordinates of bookmarked articles.
+    public var articlesCoordinates: [EventCoordinates] {
```
tyiu, line 33: Would having this class implement the `EventCoordinatesInterpreting` protocol be sufficient to replace this attribute?

Sources/NostrSDK/Events/BookmarksListEvent.swift
```diff
@@ -0,0 +1,71 @@
...
+        try super.init(kind: kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    init(content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: .bookmarksList, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    /// Ids of bookmarked kind-1 notes.
+    public var noteIds: [String] {
+        allValues(forTagName: .event) ?? []
```
tyiu, line 29: Similar to my other comment. I think it'll be good to expose relay URL information. It's difficult to discern in my head if it's useful to have both raw string ids like this as well as having a separate one that provide the full gossip information, or just expose only the one for gossip.

Sources/NostrSDK/Events/BookmarksListEvent.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+    
+    /// Coordinates of bookmarked articles.
+    public var articlesCoordinates: [EventCoordinates] {
+        tags.filter({ $0.name == TagName.eventCoordinates.rawValue }).compactMap { EventCoordinates(eventCoordinatesTag: $0) }
+    }
+    
+    /// Bookmarked links (web URLs).
+    public var links: [URL] {
```
tyiu, line 38: Would having this class implement the `ReferenceTagInterpreting` protocol be sufficient to replace this attribute?

Sources/NostrSDK/EventCreating.swift
```diff
@@ -291,6 +291,67 @@ public extension EventCreating {
...
+        for eventId in privatelyBookmarkedEventIds {
+            secretTags.append([TagName.event.rawValue, eventId])
```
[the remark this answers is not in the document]
bryanmontz: I'm not sure what you mean. For these standard lists, the secret tags get encrypted all together at the end.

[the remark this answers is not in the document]
bryanmontz: Oh I see what you mean. Yeah the NIP doesn't have any examples of adding the relay hint to the "e" tags, but I guess it could be there. I'm not sure how you would let the call site specify pairs like that unless you just made them build all the tags manually.

Sources/NostrSDK/Events/BookmarksListEvent.swift
bryanmontz: Whoops. Yeah that works. I like them having kind-specific documentation, but that probably shouldn't out-prioritize code-sharing.

Sources/NostrSDK/EventCreating.swift
bryanmontz: We have to build them into the "raw" string structure here to prepare for encryption, but maybe we could have each Tag render its own "raw" tag, which would be the array of strings. I'll work on that a bit.

Sources/NostrSDK/Events/BookmarksListEvent.swift
bryanmontz: Yeah, sorry I keep forgetting those. I think I have a compromise that reuses the code and enables custom documentation.

Sources/NostrSDK/Events/BookmarksListEvent.swift
bryanmontz: I think at the library level, it's appropriate to provide both options. I'll rethink this too.

Sources/NostrSDK/EventCreating.swift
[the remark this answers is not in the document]
tyiu: Yeah, they'd have to build them manually. But I think we can maybe hide some of that behind an abstraction by building a second layer on top that is relay aware.

bryanmontz:
Okay, @tyiu. This is ready for another look. Thanks for reviews!

Tests/NostrSDKTests/EventDecodingTests.swift
```diff
@@ -532,7 +532,7 @@ final class EventDecodingTests: XCTestCase, FixtureLoading {
         
         XCTAssertTrue(event.noteIds.contains("be8567dc210986fe5dc0ad04e02a8850b3b86e1d30c0ab674d102f0eefa68921"))
         XCTAssertTrue(event.hashtags.contains("up"))
-        XCTAssertTrue(event.links.contains(URL(string: "https://nostr.com/")!))
+        XCTAssertTrue(event.references.contains(URL(string: "https://nostr.com/")!))
```
tyiu, line 535: Should we keep this as `links` as that's the attribute that we are exposing that's closer to the language of the NIP?

Whatever we decide to do here in tests, I think we should be consistent with testing the other attribute `articlesCoordinates` / `eventCoordinates` on line 540.

Sources/NostrSDK/Tag.swift
```diff
@@ -176,6 +176,10 @@ public class Tag: Codable, Equatable {
         value == tag.value &&
         otherParameters == tag.otherParameters
     }
+    
+    var raw: [String] {
```
tyiu, line 180: We should add comments here.

Tests/NostrSDKTests/EventDecodingTests.swift
bryanmontz: Yep, it was just an oversight. Thanks!

tyiu (approved):
Looks great. Nice work!

PR #131, opened 2023-12-31
add RelayPool for connecting to and messaging with multiple relays

bryanmontz:


bryanmontz:
This PR implements a `RelayPool`, which is an object that manages a set of relays and whose interface matches that of individual relays. This makes relays and pools interchangeable.

I have already implemented changes to the demo app utilizing `RelayPool`, but I'll save those for a separate PR to simplify code review.

Closes https://github.com/nostr-sdk/nostr-sdk-ios/issues/95

Sources/NostrSDK/RelayPool.swift
```diff
@@ -0,0 +1,152 @@
...
+    private var cancellable: AnyCancellable?
+    
+    public init(relays: Set<Relay> = [], delegate: RelayDelegate? = nil) {
+        self.relays = relays
+        self.delegate = delegate
+        setUpRelays()
+    }
+    
+    public convenience init(relayURLs: Set<URL> = [], delegate: RelayDelegate? = nil) {
+        self.init(relays: Set(relayURLs.compactMap { try? Relay(url: $0) }),
```
tyiu, line 35: Should we let errors creating `Relay` bubble up?

Sources/NostrSDK/RelayPool.swift
```diff
@@ -0,0 +1,152 @@
...
+                  delegate: delegate)
+    }
+    
+    private func setUpRelays() {
+        relays.forEach { setUpRelay($0) }
+        updateEventsPublisher()
+    }
+    
+    private func setUpRelay(_ relay: Relay) {
+        relay.delegate = self
```
tyiu, line 45: Is the thought of assigning `relay.delegate` to `self` rather than `delegate` directly such that if we wanted to add additional behavior in `RelayPool`, we could?

Sources/NostrSDK/RelayPool.swift
```diff
@@ -0,0 +1,152 @@
...
+    
+    /// Subscribes to all of the relays using the supplied ``Filter``.
+    /// - Parameters:
+    ///   - filter: The filter to subscribe to.
+    ///   - subscriptionId: The subscription id. If you do not supply one, a random one will be created and returned.
+    /// - Returns: The subscription id.
+    public func subscribe(with filter: Filter, subscriptionId: String = UUID().uuidString) throws -> String {
+        do {
+            try relays.forEach { try $0.subscribe(with: filter, subscriptionId: subscriptionId) }
+        } catch {
```
tyiu, line 119: This would fail upon the first error, right? Should we do this or should we make a best effort attempt for all the relays?

Sources/NostrSDK/RelayPool.swift
```diff
@@ -0,0 +1,152 @@
...
+        }
+        return subscriptionId
+    }
+    
+    /// Attempts to close subscriptions with all relays with the supplied subscription id.
+    /// - Parameter subscriptionId: The subscription id to close subscriptions for.
+    public func closeSubscription(with subscriptionId: String) throws {
+        do {
+            try relays.forEach { try $0.closeSubscription(with: subscriptionId) }
+        } catch {
```
tyiu, line 130: Same here.

Sources/NostrSDK/RelayPool.swift
```diff
@@ -0,0 +1,152 @@
...
+            logger.error("An error occurred while closing a subscription with subscription id: \(subscriptionId): \(error)")
+        }
+    }
+    
+    /// Publishes an event to the relays.
+    /// - Parameter event: The ``NostrEvent`` to publish
+    public func publishEvent(_ event: NostrEvent) throws {
+        try relays.forEach {
```
tyiu, line 138: Same here. Should it be best effort try all the relays?

tyiu (commented):
Looks great overall! Left some comments.

Sources/NostrSDK/RelayPool.swift
bryanmontz: Yeah, this makes sense. It makes for a not-ideal callsite when using SwiftUI and `@StateObject` with a pool though.

Sources/NostrSDK/RelayPool.swift
bryanmontz: Hmm. Actually I didn't even think of doing it that way. But it's functionally equivalent to what I did and simpler. I'll do that, thanks.

Sources/NostrSDK/RelayPool.swift
bryanmontz: Yeah. Thinking about how to solve this.

bryanmontz: Okay, I made this update, and I think it's probably best. However it makes It so that this function no longer `throws`, so the caller won't be notified of errors. It'll still produce messages in the console though, and they're better now because they contain the relay URL.

tyiu: That's true. We can iterate on this over time. Maybe it could make sense to have two approaches exposed. It might still be valuable to bubble up errors in a first-class way while still attempting a best-effort subscription on the relays, and let applications decide how they want to deal with errors.

This is fine for now, though. We can always iterate on it.

tyiu (approved):

PR #133, opened 2024-01-07
update Demo app to support RelayPool

bryanmontz:
This PR updates the demo app from using a single relay to using a RelayPool. For now it hardcodes three relays and connects to all of them. For now, relays can be removed while the app is running. 

It also adds a new view for managing the set of relays and reconfigures the initial view to provide a user path to the new view.

|Before|After|
|---|---|
|drawing|drawing|

Still working:
drawing

Potential Future Improvements
In the future we could add support for adding new relays while the app is running and caching relay information so that it doesn't reset every time you launch the app. These weren't necessary to demonstrate the functionality of the relay pool, in my opinion.


tyiu (approved):
Looks good to me. Thanks, @bryanmontz!

PR #134, opened 2024-01-08
improve the Query Relays view

bryanmontz:
Adds some options and clarity to the kind picker on the Query Relays view.

|Before|After|
|---|---|
|Simulator Screenshot - iPhone 15 Pro - 2024-01-08 at 07 05 18|Simulator Screenshot - iPhone 15 Pro - 2024-01-08 at 07 01 56|

https://github.com/nostr-sdk/nostr-sdk-ios/assets/445882/5b59b782-e0c5-4309-9b3f-d71baf784952


demo/NostrSDKDemo/Demo Views/QueryRelayDemoView.swift
```diff
@@ -32,8 +42,12 @@ struct QueryRelayDemoView: View {
                 }
 
                 Picker("Kind", selection: $selectedKind) {
-                    ForEach(kindOptions, id: \.self) { number in
-                        Text("\(number)")
+                    ForEach(kindOptions.keys.sorted(), id: \.self) { number in
+                        if let name = kindOptions[number] {
+                            Text("\(name) (\(number))")
```
tyiu, line 47: It's a bit unusual to display kind numbers formatted (with separators) as it's formatted differently depending on the locale, and the NIPs repo doesn't do that.

Perhaps something like this would be better?

```swift
if let name = kindOptions[number] {
    Text("\(name) (\(String(number)))")
} else {
    Text(String(number))
}
```

tyiu (approved):
Looks great overall! Left a minor nitpick suggestion. Approved the PR. Feel free to address the feedback and merge at your own discretion.

demo/NostrSDKDemo/Demo Views/QueryRelayDemoView.swift
bryanmontz: Agree, thanks.

PR #136, opened 2024-01-12
Remove label cruft in NIP-52 to reflect changes to spec

tyiu:
Reflects the changes from this merged PR in the NIPs repo: https://github.com/nostr-protocol/nips/pull/980

### DIFF
--- a/Sources/NostrSDK/EventCreating.swift
+++ b/Sources/NostrSDK/EventCreating.swift
@@ -612,13 +612,11 @@ public extension EventCreating {
         var tags: [Tag] = [
             calendarEventCoordinates.tag,
             Tag(name: .identifier, value: identifier),
-            Tag(name: .labelNamespace, value: "status"),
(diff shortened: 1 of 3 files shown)


Sources/NostrSDK/EventCreating.swift
```diff
@@ -612,13 +612,11 @@ public extension EventCreating {
         var tags: [Tag] = [
             calendarEventCoordinates.tag,
             Tag(name: .identifier, value: identifier),
-            Tag(name: .labelNamespace, value: "status"),
-            Tag(name: .label, value: status.rawValue, otherParameters: ["status"])
+            Tag(name: "status", value: status.rawValue)
```
tyiu, line 615: I chose not to standardize the `status` and `fb` tags in the `TagName` enum as their usages are constrained to only calendar event RSVPs and it seems that we want kind-specific tag enums somehow if that's possible, if we figure that out down the line.

bryanmontz (approved):

PR #140, opened 2024-04-05
Replace Sodium with CryptoSwift

tyiu:
Using `Sodium`, which is backed by a C library, broke documentation generation for some reason. We were using it for the `ChaCha20` cipher for NIP-44 encryption, introduced in this commit: https://github.com/nostr-sdk/nostr-sdk-ios/commit/f1384b7a447a3d16cf97a93141d8a2001e884c74

This PR replaces `Sodium` with `CryptoSwift`, a pure Swift library, which implements the ChaCha20 cipher and seems to fix documentation generation. I tried it both locally and on the `main` remote branch on my fork.

The fact that `NIP44v2EncryptingTests` passes gives me confidence that the `CryptoSwift` usage is functionally equivalent to what we were doing with `Sodium`.

### DIFF
--- a/Sources/NostrSDK/NIP44v2Encrypting.swift
+++ b/Sources/NostrSDK/NIP44v2Encrypting.swift
@@ -6,8 +6,8 @@
 //
 
 import Foundation
-import Clibsodium
 import CryptoKit
+import CryptoSwift
 import secp256k1
 
 public enum NIP44v2EncryptingError: Error {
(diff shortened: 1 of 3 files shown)


Sources/NostrSDK/NIP44v2Encrypting.swift
```diff
@@ -199,16 +199,21 @@ extension NIP44v2Encrypting {
 
         let combined = aad + message
 
-        return Data(HMAC<CryptoKit.SHA256>.authenticationCode(for: combined, using: SymmetricKey(data: key)).bytes)
+        return Data(CryptoKit.HMAC<CryptoKit.SHA256>.authenticationCode(for: combined, using: SymmetricKey(data: key)))
+    }
+
+    private func toBytes(from data: Data) -> [UInt8] {
```
tyiu, line 205: This is kind of gross, but I'm doing this intentionally because I was getting ambiguous definition clashes from importing the `CryptoSwift` library. It defines a `var bytes: [UInt8]` as an extension to `Data`, and it clashes with `bytes` from the `secp256k1` library. I chose to copy the implementation here locally and use this instead.

bryanmontz (approved):
I'm way out of my depth here. But I love the fact that you changed the guts and didn't have to touch the related tests. 👍

PR #144, opened 2024-04-15
Reorganize Nostr event code to be grouped by event kind

tyiu:
We have one massive `EventCreating.swift` file that contains all the Nostr event creation functions. That approach is not scalable because the file will grow without bound as we support more event kinds. Moreover, it becomes increasingly difficult to maintain that code.

Similarly, we have massive `EventCreatingTests.swift` and `EventDecodingTests.swift` test files which also grow without bound and are difficult to maintain.

I've reorganized the code so that the `EventCreating` protocol function(s) that create a specific Nostr event kind will be located in the same file as where the Nostr event kind class is defined. Moreover, the tests for each Nostr event kind were reorganized such that the creation and decoding is located in the same test file, specific to that Nostr event kind.

There is no change in any functionality or public function contracts The only changes were:
1. Where the functions are located, and
2. Naming of some test functions to distinguish event creating tests from event decoding tests in the same test file.


Tests/NostrSDKTests/Events/DeletionEventTests.swift
```diff
@@ -0,0 +1,59 @@
...
+//  DeletionEventTests.swift
+//  
+//
+//  Created by Terry Yiu on 4/14/24.
+//
+
+@testable import NostrSDK
+import XCTest
+
+final class DeletionEventTests: XCTestCase, EventCreating, EventVerifying, FixtureLoading {
```
tyiu, line 11: I noticed that there's a missing test for decoding deletion events. We can add that separately.

bryanmontz (approved):
This makes a lot of sense and will help the SDK scale better. Thanks! And thanks for not mixing in any functional changes; that limits the scope and makes it much easier to reason about.

PR #149, opened 2024-04-20
Make NostrEvent conform to the Hashable protocol and change its Equatable implementation to compare against all its properties

tyiu:
This change enables NostrEvent and its subclasses to be hashable, which is great for allowing indexing and deduplicating in Set and Dictionary data structures.

This change also prevents events with the same `id` from being erroneously equated the same as another event with the same `id` if any of its other properties are different. Note that this is a breaking change.

### DIFF
--- a/Sources/NostrSDK/Events/NostrEvent.swift
+++ b/Sources/NostrSDK/Events/NostrEvent.swift
@@ -12,7 +12,13 @@ import Foundation
 /// > Note: [NIP-01 Specification](https://github.com/nostr-protocol/nips/blob/master/01.md#events-and-signatures)
 public class NostrEvent: Codable, Equatable, Hashable {
     public static func == (lhs: NostrEvent, rhs: NostrEvent) -> Bool {
-        lhs.isEqual(to: rhs)
+        lhs.id == rhs.id &&
+        lhs.pubkey == rhs.pubkey &&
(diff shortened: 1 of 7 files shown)


Sources/NostrSDK/Events/NostrEvent.swift
```diff
@@ -114,4 +124,14 @@ public class NostrEvent: Codable, Equatable {
     public func allValues(forTagName tag: TagName) -> [String] {
         tags.filter { $0.name == tag.rawValue }.map { $0.value }
     }
+
+    func isEqual(to nostrEvent: NostrEvent) -> Bool {
```
bryanmontz, line 128: I'd rather not expose two different methods of checking equality, and the `isEqual(to:)` pattern feels like a lot of Apple's legacy Objective-C stuff. Can you just move all of this function into the `==`?

bryanmontz (changes requested):

Sources/NostrSDK/Events/NostrEvent.swift
tyiu: Great feedback. I thought it was weird but tried to be consistent since other classes followed this pattern. I've moved it into `==`, and I've done the same for other classes as well.

bryanmontz (approved):
Great call to update all the other similar spots. I'm not sure why we did that.

PR #150, opened 2024-04-20
Remove unused files

dcadenas:
Removed the `NostrSDK` struct, which shared its name with the library, to resolve a naming conflict. This change enables unambiguous access to types through fully qualified paths, such as `NostrSDK.PublicKey`.

This simplifies namespace usage when managing multiple dependencies with similar naming conventions.


bryanmontz (approved):
Yeah, good call. We were using this as a placeholder for future configuration, but we can use a different name. Thanks for the contribution!

PR #151, opened 2024-05-02
Relay request and response additions

tyiu:
Please review this PR on a per-commit basis. Otherwise, it will be difficult to review.
- Reorder relay request and relay response enums to match the listed ordering in the specs for easier maintainability
- Fix `RelayResponse` to not drop human readable messages for `OK` messages
- Add `CLOSED` relay response
- Add `AUTH` relay request and response

This PR brings `RelayRequest` and `RelayResponse` to completion per [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md), [NIP-42 - Auth](https://github.com/nostr-protocol/nips/blob/master/42.md), and [NIP-45 - Event counts](https://github.com/nostr-protocol/nips/blob/master/45.md). The implementation was previously incomplete, covering only part of the spec.

Although this PR adds AUTH, there are no functions to actually authenticate on the `RelayOperating` protocol yet. That will be added as a separate PR.
This issue tracks the `AUTH` work: https://github.com/nostr-sdk/nostr-sdk-ios/issues/21


tyiu:
There's something wrong with the CI builds. I'm getting flaky results or they're breaking in a way that's unexpected. It seems that my PR shouldn't affect them, but maybe I'm missing something.

tyiu:
> There's something wrong with the CI builds. I'm getting flaky results or they're breaking in a way that's unexpected. It seems that my PR shouldn't affect them, but maybe I'm missing something.

It's not my PR. I tried the build on a different branch that's basically just the current `main` branch and it fails in the same way for the Swift 5.7.1 build. Either the tooling or environment has changed.

tyiu:
> > There's something wrong with the CI builds. I'm getting flaky results or they're breaking in a way that's unexpected. It seems that my PR shouldn't affect them, but maybe I'm missing something.
> 
> It's not my PR. I tried the build on a different branch that's basically just the current `main` branch and it fails in the same way for the Swift 5.7.1 build. Either the tooling or environment has changed.
> 
> ```
> Run swift build
>   swift build
>   swift test
>   shell: /bin/bash -e {0}
>   env:
>     TOOLCHAINS: org.swift.571202211011a
> error: 'nostr-sdk-ios': Invalid manifest
> error: compile command failed due to signal 6 (use -v to see invocation)
> /Applications/Xcode_15.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk/usr/lib/swift/Dispatch.swiftmodule/arm64e-apple-macos.swiftinterface:84:20: error: no type named 'DispatchQueue' in module 'Dispatch'
> extension Dispatch.DispatchQueue {
>                    ^
> ```

https://github.com/nostr-sdk/nostr-sdk-ios/pull/152 fixes the build issue.

Sources/NostrSDK/Events/AuthenticationEvent.swift
```diff
@@ -0,0 +1,55 @@
...
+    @available(*, unavailable, message: "This initializer is unavailable for this class.")
+    override init(kind: EventKind, content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+
+    init(content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: .authentication, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+
+    /// The relay URL where this event authenticates to it.
```
bryanmontz, line 28:
```suggestion
    /// The relay URL this event authenticates to.
```

Sources/NostrSDK/RelayResponse.swift
```diff
@@ -29,49 +29,55 @@ enum RelayResponse: Decodable {
...
+        case closed = "CLOSED"
+        case notice = "NOTICE"
         case auth = "AUTH"
+        case count = "COUNT"
     }
     
-    struct OKMessage {
-        let type: OKMessageType
-        let message: String?
+    struct RelayResponseMessage {
```
bryanmontz, line 40: Because this struct is defined inside the scope of `RelayResponse` you can simply call this one `Message`. Then external to the `RelayResponse` type, this struct will be exposed as `RelayResponse.Message`. The same would be true of the prefix if you defined it inside this `Message` type. You could call it simply `Prefix`, and it would then be `RelayResponse.Message.Prefix`. This is more in alignment with how Apple defines its own types.

Tests/NostrSDKTests/Fixtures/ok_unknown_reason.json
```diff
@@ -0,0 +1 @@
+["OK", "b1a649ebe8b435ec71d3784793f3bbf4b93e64e17568a741aecd4c7ddeafce30", true, "unknown: reason: unknown"]
```
bryanmontz, line 1: Is it possible that the message will have two colons in it?

Tests/NostrSDKTests/RelayResponseDecodingTests.swift
```diff
@@ -99,23 +105,52 @@ final class RelayResponseDecodingTests: XCTestCase, FixtureLoading {
...
+            XCTAssertEqual(subscriptionId, "some-subscription-id")
+        } else {
+            XCTFail("failed to decode")
+        }
+    }
+
+    func testDecodeClosedMessage() throws {
+        let data = try loadFixtureData("closed")
+
+        if let relayResponse = RelayResponse.decode(data: data) {
```
bryanmontz, line 132: Suggestion: I see that this was the existing pattern in this file, and keeping that is fine. In general, however, it is cleaner to use `XCTUnwrap` rather than `if let`/`guard let` in tests.

bryanmontz (changes requested):
Thanks, Terry!

Tests/NostrSDKTests/Fixtures/ok_unknown_reason.json
tyiu: Yes, it's possible. Technically the relay can give us whatever it wants and not conform to NIP-01. I wanted to stress test that possibility.

Sources/NostrSDK/RelayResponse.swift
tyiu: Nice feedback. Fixed!

Tests/NostrSDKTests/RelayResponseDecodingTests.swift
tyiu: I cleaned up all those usages in this test file. Thanks!

bryanmontz (approved):

PR #155, opened 2024-05-14
Change properties to public from internal, so that this public struct…

PavleKreator:
… is actually useful.


tyiu (approved):
Looks good. Thanks for the fix!

PR #158, opened 2024-05-21
Change NIP-04 direct message and encryption naming to avoid ambiguity with NIP-17

tyiu:
These two NIPs overload the term "direct message", so we need to use the term "legacy" for the old one to differentiate them.
[NIP-04 - Encrypted Direct Messages](https://github.com/nostr-protocol/nips/blob/master/04.md)
[NIP-17 - Private Direct Messages](https://github.com/nostr-protocol/nips/blob/master/17.md)

Note that this PR contains breaking changes.


demo/NostrSDKDemo/Demo Views/QueryRelayDemoView.swift
```diff
@@ -97,7 +97,7 @@ struct QueryRelayDemoView: View {
         } else {
             authors = [authorPubkey]
         }
-        return Filter(authors: authors, kinds: [selectedKind])
+        return Filter(authors: authors, kinds: [selectedKind])!
```
tyiu, line 100: Unrelated fix for the demo app.

The `Filter` initializer was [changed](https://github.com/nostr-sdk/nostr-sdk-ios/pull/154) to be able to return `nil` if `tags` contains tag names that are not in the basic Latin alphabet of A-Z or a-z. However, we're not using tags, so it will never be `nil`. I recognize that forcing the unwrap of an optional is bad, but since it's just a demo app and we know that it will always be non-nil, I thought it would be ok.

Sources/NostrSDK/EventKind.swift
```diff
@@ -30,8 +30,9 @@ public enum EventKind: RawRepresentable, CaseIterable, Codable, Equatable, Hasha
 
     /// This kind of event should have a recipient pubkey tag.
     ///
-    /// See [NIP-04 - Direct Messages](https://github.com/nostr-protocol/nips/blob/master/04.md)
-    case directMessage
+    /// See [NIP-04 - Encrypted Direct Message](https://github.com/nostr-protocol/nips/blob/master/04.md)
+    /// > Warning: Deprecated in favor of [NIP-17 - Private Direct Messages](https://github.com/nostr-protocol/nips/blob/master/04.md).
+    case nip04EncryptedDirectMessage
```
bryanmontz, line 35: I get that we need to distinguish between types of direct messaging, but--and this is just a suggestion, you have more context than I do--what if we called the nip04 ones "legacy"? So it would be obvious to the client that this is "the old way". We could also mark it `@deprecated`.

Sources/NostrSDK/Events/NIP04EncryptedDirectMessageEvent.swift
```diff
@@ -9,8 +9,9 @@ import Foundation
 
 /// An event that contains an encrypted message.
 ///
-/// > Note: [NIP-04 Specification](https://github.com/nostr-protocol/nips/blob/master/04.md)
-public final class DirectMessageEvent: NostrEvent, DirectMessageEncrypting {
+/// > Note: [NIP-04 - Encrypted Direct Message](https://github.com/nostr-protocol/nips/blob/master/04.md)
+/// > Warning: Deprecated in favor of [NIP-17 - Private Direct Messages](https://github.com/nostr-protocol/nips/blob/master/04.md).
+public final class NIP04EncryptedDirectMessageEvent: NostrEvent, NIP04DirectMessageEncrypting {
```
bryanmontz, line 14: Let's mark this all this stuff as `@deprecated`.

demo/NostrSDKDemo/Demo Views/QueryRelayDemoView.swift
bryanmontz: This change is fine, but it pollutes this PR and the resulting commit. If we were ever to need to revert the PR's squashed commit, it would revert this unrelated thing also and we might not notice it.

bryanmontz (changes requested):

demo/NostrSDKDemo/Demo Views/QueryRelayDemoView.swift
tyiu: Fair point. I got lazy. I'm happy to split this out into a separate PR.

Sources/NostrSDK/EventKind.swift
tyiu: I think that's a fine suggestion.

There's one edge case I can think of, though. If the new NIP-17 DM kind gets deprecated in the future (not that I'm anticipating it anytime soon, but anything is possible in the Nostr world), we would have another "legacy" DM kind, making the "legacy" naming convention overloaded.

However, I think optimizing for developer ergonomics might make more sense here. I'll make that change.

bryanmontz: Right, it's one possible future to have many DM kinds as the protocol progresses. In that case though, I would recommend that the current best practice one would always be named the most simply, like you said for developer ergonomics. And then all the legacy types would have uglier names with sufficient distinction, such as `NIP04LegacyDirectMessageEvent`.

bryanmontz (approved):

PR #164, opened 2024-06-26
Add function to RelayDelegate to receive all relay response types

tyiu:
The main purpose of this change is that there is currently no way for consumers of the library to reasonably manage their relay subscriptions.

For example, if they wanted to subscribe to events up to only what is stored on the relay, but they do not want to receive any *new* live events that come in, they will want to close their subscription after receiving an `EOSE` (end of stored events) response rather than keeping the subscription perpetually open.

By allowing folks to listen on all types of relay responses (not just ones with Nostr events), it provides more flexibility to the developer in maintaining the state of their application.

- [ ] Do not merge until https://github.com/nostr-sdk/nostr-sdk-ios/pull/162 gets merged
- [ ] Change this PR base branch to be `main` after the above PR is merged


Sources/NostrSDK/Relay.swift
```diff
@@ -51,13 +51,17 @@ public protocol RelayOperating {
...
 }
 
-/// An optional interface for receiving state updates and events
+/// An optional interface for receiving state updates and responses from relays.
 public protocol RelayDelegate: AnyObject {
     func relayStateDidChange(_ relay: Relay, state: Relay.State)
+
+    func relay(_ relay: Relay, didReceive response: RelayResponse)
+
+    @available(*, deprecated, message: "`relay(Relay, didReceive: RelayEvent)` was replaced by `relay(Relay, didReceive: RelayResponse)` and will be removed shortly.")
```
tyiu, line 60: Thoughts on if deprecating this function is the right thing to do?

bryanmontz: Maybe we could keep it as not deprecated if there is ever a reason to just need the `RelayEvent` and it's a simpler object to interact with, but I don't have a strong opinion about it.

Tests/NostrSDKTests/RelayTests.swift
```diff
@@ -93,8 +104,10 @@ final class RelayTests: XCTestCase {
         let filter = try XCTUnwrap(Filter(kinds: [1], limit: 1))
         let subscriptionId = try relay.subscribe(with: filter)
         
-        wait(for: [receiveExpectation!], timeout: 10)
-        
+        wait(for: [receiveResponseEventExpectation!], timeout: 10)
```
bryanmontz, line 107: Since `wait(for:timeout:)` takes an array of expectations, you could just put them all into one wait.

bryanmontz (approved):
Gave some suggestions.

PR #165, opened 2024-06-28
Add missing fields to UserMetadataEvent and enable MetadataEvents to be updated without wiping out existing fields

tyiu:


Sources/NostrSDK/Events/MetadataEvent.swift
```diff
@@ -8,48 +8,66 @@
...
     public let bannerPictureURL: URL?
-    
+
+    /// A boolean to clarify that the content is entirely or partially the result of automation, such as with chatbots or newsfeeds.
+    /// > Note: [NIP-24 Extra metadata fields and tags](https://github.com/nostr-protocol/nips/blob/master/24.md#kind-0)
+    public let bot: Bool?
+
+    /// The user's LUD-06 Lightning URL (LNURL).
+    /// > Note: [NIP-57 Lightning Zaps](https://github.com/nostr-protocol/nips/blob/master/57.md#protocol-flow)
+    public let lud06: String?
```
bryanmontz, line 46: Suggestion: These names are so cryptic to me. I know it's what's in the spec, but it's hard to understand. What if we called this `lightningURLString` and the lud16 `lightningAddress`? Yes, it would take a lookup into documentation to understand the field it's actually tied to, but it would be much more understandable for a client dev. Up to you though.

Sources/NostrSDK/Events/MetadataEvent.swift
```diff
@@ -8,48 +8,66 @@
...
     public let pictureURL: URL?
     
     /// A URL to retrieve the user's banner image.
+    /// > Note: [NIP-24 Extra metadata fields and tags](https://github.com/nostr-protocol/nips/blob/master/24.md#kind-0)
     public let bannerPictureURL: URL?
-    
+
+    /// A boolean to clarify that the content is entirely or partially the result of automation, such as with chatbots or newsfeeds.
+    /// > Note: [NIP-24 Extra metadata fields and tags](https://github.com/nostr-protocol/nips/blob/master/24.md#kind-0)
+    public let bot: Bool?
```
bryanmontz, line 42: `isBot`? With it just `bot` it sounds like the datatype might be a model object that describes a bot, rather than a bool.

bryanmontz (approved):
Just a couple naming suggestions, but take 'em or leave 'em. Thanks, @tyiu.

PR #166, opened 2024-06-30
Add NIP-19 bech32 identifier generation for note, nevent, and naddr prefixes

tyiu:
https://github.com/nostr-protocol/nips/blob/master/19.md

This change enables clients to generate shareable identifiers easily for human consumption.

I also refactored the `npub` and `nsec` bech32 prefixes into the `Bech32IdentifierType` enum.

### DIFF
--- a/Sources/NostrSDK/Bech32IdentifierType.swift
+++ b/Sources/NostrSDK/Bech32IdentifierType.swift
@@ -1,0 +1,19 @@
+//
+//  Bech32IdentifierType.swift
+//
+//
+//  Created by Terry Yiu on 6/30/24.
+//
+
+/// The type of Bech32-encoded identifier.
+/// These identifiers can be used to succinctly encapsulate metadata to aid in the discovery of events and users.
+/// See [NIP-19](https://github.com/nostr-protocol/nips/blob/master/19.md) for information about how these Bech32-encoded
+public enum Bech32IdentifierType: String {
+    case publicKey = "npub"
+    case privateKey = "nsec"
+    case note = "note"
+    case profile = "nprofile"
+    case event = "nevent"
+    case relay = "nrelay"
+    case address = "naddr"
+}
--- a/Sources/NostrSDK/Bech32IdentifierType.swift
+++ b/Sources/NostrSDK/Bech32IdentifierType.swift
@@ -7,7 +7,8 @@
 
 /// The type of Bech32-encoded identifier.
 /// These identifiers can be used to succinctly encapsulate metadata to aid in the discovery of events and users.
-/// See [NIP-19](https://github.com/nostr-protocol/nips/blob/master/19.md) for information about how these Bech32-encoded
+/// See [NIP-19](https://github.com/nostr-protocol/nips/blob/master/19.md) for information about how these
+/// identifiers are encoded and used.
 public enum Bech32IdentifierType: String {
     case publicKey = "npub"
     case privateKey = "nsec"
--- a/Sources/NostrSDK/Events/NonParameterizedReplaceableEvent.swift
+++ b/Sources/NostrSDK/Events/NonParameterizedReplaceableEvent.swift
@@ -17,16 +17,16 @@ public extension NonParameterizedReplaceableEvent {
         return try? EventCoordinates(kind: kind, pubkey: publicKey, relayURL: relayURL)
     }
 
-    func shareableEventCoordinates(relayURLStrings: [String]? = nil, excludeAuthor: Bool = false, excludeKind: Bool = false) throws -> String {
+    func shareableEventCoordinates(relayURLStrings: [String]? = nil, includeAuthor: Bool = true, includeKind: Bool = true) throws -> String {
         let validatedRelayURLStrings = try relayURLStrings?.map {
             try validateRelayURLString($0)
         }.map { $0.absoluteString }
 
         var metadata = Metadata(relays: validatedRelayURLStrings, identifier: "")
-        if !excludeAuthor {
+        if includeAuthor {
             metadata.pubkey = pubkey
         }
-        if !excludeKind {
+        if includeKind {
             metadata.kind = UInt32(kind.rawValue)
         }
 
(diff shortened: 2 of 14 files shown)


Sources/NostrSDK/Bech32IdentifierType.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  Bech32IdentifierType.swift
+//
+//
+//  Created by Terry Yiu on 6/30/24.
+//
+
+/// The type of Bech32-encoded identifier.
+/// These identifiers can be used to succinctly encapsulate metadata to aid in the discovery of events and users.
+/// See [NIP-19](https://github.com/nostr-protocol/nips/blob/master/19.md) for information about how these Bech32-encoded
```
bryanmontz, line 10: I see that this was copied from _MetadataCoding.swift_, but the end of this line doesn't finish the sentence.

Sources/NostrSDK/Events/NonParameterizedReplaceableEvent.swift
```diff
@@ -16,4 +16,20 @@ public extension NonParameterizedReplaceableEvent {
 
         return try? EventCoordinates(kind: kind, pubkey: publicKey, relayURL: relayURL)
     }
+
+    func shareableEventCoordinates(relayURLStrings: [String]? = nil, excludeAuthor: Bool = false, excludeKind: Bool = false) throws -> String {
```
bryanmontz, line 20: Generally it is less cognitive load for `Bool`s to be positively stated: `includeAuthor`, `includeKind`. That way you don't have to deal with double-negative like you would with `!excludeAuthor`.

Sources/NostrSDK/Events/NostrEvent.swift
```diff
@@ -141,3 +141,47 @@ public class NostrEvent: Codable, Equatable, Hashable {
...
+    /// It is not meant to be used inside the standard NIP-01 event formats or inside the filters.
+    /// They are meant for human-friendlier display and input only.
+    /// Clients should still accept keys in both hex and npub format and convert internally.
+    ///
+    /// > Note: [NIP-19 bech32-encoded entities](https://github.com/nostr-protocol/nips/blob/master/19.md)
+    public var bech32NoteId: String? {
+        guard let data = id.hexDecoded else {
+            return nil
+        }
+        return Bech32.encode(NostrEvent.bech32NoteIdPrefix, baseEightData: data)
```
bryanmontz, line 158: Rather than hard-coding a second source of this information, why not use `Bech32IdentifierType.note.rawValue`?
```suggestion
        return Bech32.encode(Bech32IdentifierType.note.rawValue, baseEightData: data)
```

Sources/NostrSDK/Events/NostrEvent.swift
```diff
@@ -141,3 +141,47 @@ public class NostrEvent: Codable, Equatable, Hashable {
...
+    }
+
+    /// Gets a shareable human-interactable event identifier for this event.
+    /// The identifier is bech32-formatted with a prefix of `nevent` using a binary-encoded list of TLV (type-length-value).
+    /// The identifier have all the information needed for the event to be found, which includes the
```
bryanmontz, line 163:
```suggestion
    /// The identifier has all the information needed for the event to be found, which includes the
```

Sources/NostrSDK/Events/NostrEvent.swift
```diff
@@ -141,3 +141,47 @@ public class NostrEvent: Codable, Equatable, Hashable {
...
+    /// event id, optionally the relays, optionally the author's public key, and optionally the event kind number.
+    /// - Parameters:
+    ///   - relayURLs: The String representations of relay URLs in which the event is more likely to be found, encoded as ASCII.
+    ///   - excludeAuthor: Whether the author public key should be excluded from the identifier.
+    ///   - excludeKind: Whether the event kind number should be excluded from the identifier.
+    /// - Throws: `URLError.Code.badURL`, `RelayURLError.invalidScheme`, `TLVCodingError.failedToEncode`
+    ///
+    /// > Note: [NIP-19 bech32-encoded entities](https://github.com/nostr-protocol/nips/blob/master/19.md)
+    public func shareableEventIdentifier(relayURLStrings: [String]? = nil, excludeAuthor: Bool = false, excludeKind: Bool = false) throws -> String {
```
bryanmontz, line 172: Same comment here about stating `Bool`s positively.

Sources/NostrSDK/Events/ParameterizedReplaceableEvent.swift
```diff
@@ -22,4 +22,20 @@ public extension ParameterizedReplaceableEvent {
 
         return try? EventCoordinates(kind: kind, pubkey: publicKey, identifier: identifier, relayURL: relayURL)
     }
+
+    func shareableEventCoordinates(relayURLStrings: [String]? = nil, excludeAuthor: Bool = false, excludeKind: Bool = false) throws -> String {
+        let validatedRelayURLStrings = try relayURLStrings?.map {
```
bryanmontz, line 27: Since this block of code is nearly identifier between parameterized and non-parameterized replaceable events, you could factor out this code into a helper function inside an extension of the `ReplaceableEvent` protocol. It would need to accept one more argument for `identifier`.

That way, here (and without the identifier in the non-parameterized variant) you could simply call:
```swift
try shareableEventCoordinates(relayURLStrings: relayURLStrings,
                                      excludeAuthor: excludeAuthor,
                                      excludeKind: excludeKind,
                                      identifier: identifier ?? "")
```

bryanmontz (changes requested):
Great work! Thanks, @tyiu.

Sources/NostrSDK/Events/NonParameterizedReplaceableEvent.swift
tyiu: That's good feedback. My brain is hardwired to always have the default parameters be false, which is why I went on the `exclude` route. But doing the inverse makes more sense in this case.

Sources/NostrSDK/Events/NostrEvent.swift
tyiu: You're right, I missed it. Nice catch!

bryanmontz (approved):

PR #169, opened 2024-07-14
Add support for NIP-65 Relay List Metadata

tyiu:
[NIP-65 Relay List Metadata](https://github.com/nostr-protocol/nips/blob/master/65.md)

### DIFF
--- a/Sources/NostrSDK/Events/RelayListMetadataEvent.swift
+++ b/Sources/NostrSDK/Events/RelayListMetadataEvent.swift
@@ -95,7 +95,7 @@ public struct UserRelayMetadata: Equatable {
     /// A `nil` value is returned if the relay URL string is invalid.
     ///
     /// > Note: [NIP 65 - Relay List Metadata](https://github.com/nostr-protocol/nips/blob/master/65.md)
-    public init?(relayURL: URL, marker: Marker) {
+    public init?(relayURL: URL, marker: Marker = .readAndWrite) {
         guard let validatedRelayURL = try? RelayURLValidator.shared.validateRelayURL(relayURL) else {
             return nil
         }
@@ -131,7 +131,7 @@ public extension EventCreating {
             if let existingMetadata = deduplicatedMetadata[metadata.relayURL] {
                 // If the user relay metadata is identical between the duplicates,
                 // or if the existing one already has a read and write marker, skip it.
-                guard existingMetadata.marker != metadata.marker && existingMetadata.marker != .readAndWrite else {
+                if existingMetadata.marker == metadata.marker || existingMetadata.marker == .readAndWrite {
                     continue
                 }
 
--- a/Sources/NostrSDK/Events/RelayListMetadataEvent.swift
+++ b/Sources/NostrSDK/Events/RelayListMetadataEvent.swift
@@ -1,0 +1,154 @@
+//
+//  RelayListMetadataEvent.swift
+//
+//
+//  Created by Terry Yiu on 7/13/24.
+//
+
+import Foundation
+import OrderedCollections
+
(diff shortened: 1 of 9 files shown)


Sources/NostrSDK/Events/RelayListMetadataEvent.swift
```diff
@@ -0,0 +1,154 @@
...
+
+        self.relayURL = relayURL
+    }
+
+    /// Creates a ``UserRelayMetadata`` from a relay ``URL`` and ``Marker``.
+    ///
+    /// A `nil` value is returned if the relay URL string is invalid.
+    ///
+    /// > Note: [NIP 65 - Relay List Metadata](https://github.com/nostr-protocol/nips/blob/master/65.md)
+    public init?(relayURL: URL, marker: Marker) {
```
bryanmontz, line 98: Just a suggestion: What if you made the marker have a default value of `.readAndWrite`? That way callers in most cases wouldn't need to pass it at all.

```suggestion
    public init?(relayURL: URL, marker: Marker = .readAndWrite) {
```

Sources/NostrSDK/Events/RelayListMetadataEvent.swift
```diff
@@ -0,0 +1,154 @@
...
+        // Using an ordered dictionary to retain the order of the list while de-duplicating the data.
+        var deduplicatedMetadata = OrderedDictionary<URL, UserRelayMetadata>()
+
+        for metadata in relayMetadataList {
+            if let existingMetadata = deduplicatedMetadata[metadata.relayURL] {
+                // If the user relay metadata is identical between the duplicates,
+                // or if the existing one already has a read and write marker, skip it.
+                guard existingMetadata.marker != metadata.marker && existingMetadata.marker != .readAndWrite else {
+                    continue
+                }
```
bryanmontz, line 136: I would recommend using `if` rather than `guard` here. `guard` is better for protecting the last line in a function from all the undesirable cases. This is difficult to read here. And since you have a comment above it that uses the word `if`, it would be easier to check if the code matches the comment if you use `if`.

```suggestion
                if existingMetadata.marker == metadata.marker || existingMetadata.marker == .readAndWrite {
                    continue
                }
```

Sources/NostrSDK/Events/RelayListMetadataEvent.swift
```diff
@@ -0,0 +1,154 @@
...
+
+                // Any other permutation of markers will result in a combined marker of read and write.
+                switch metadata.marker {
```
bryanmontz, line 139: A simple `if`-`else` would suffice here.

Tests/NostrSDKTests/Events/RelayListMetadataEventTests.swift
```diff
@@ -0,0 +1,140 @@
...
+        XCTAssertEqual(event.relayMetadataList[0].marker, .readAndWrite)
+        XCTAssertEqual(event.relayMetadataList[0].tag, tag1)
+        XCTAssertEqual(event.relayMetadataList[1].relayURL.absoluteString, "wss://relay.primal.net/")
+        XCTAssertEqual(event.relayMetadataList[1].marker, .read)
+        XCTAssertEqual(event.relayMetadataList[1].tag, tag2)
+        XCTAssertEqual(event.relayMetadataList[2].relayURL.absoluteString, "wss://relay.nostr.band/")
+        XCTAssertEqual(event.relayMetadataList[2].marker, .read)
+        XCTAssertEqual(event.relayMetadataList[2].tag, tag3)
+    }
+
```
bryanmontz, line 139: Would it be good to test decoding malformed data? Maybe a relay puts an invalid url in. Or maybe they put something other than "read" or "write" where the marker should be.

bryanmontz (approved):
Awesome work as usual, @tyiu! Approve to merge, but I left a few suggestions.

PR #178, opened 2024-09-12
Adds support for NIP-28

masterial:
[NIP-28 Public Chat](https://github.com/nostr-protocol/nips/blob/master/28.md)

### DIFF
--- a/Sources/NostrSDK/Events/Channels/CreateChannelEvent.swift
+++ b/Sources/NostrSDK/Events/Channels/CreateChannelEvent.swift
@@ -1,0 +1,47 @@
+//
+//  CreateChannelEvent.swift
+//
+//
+//  Created by Konstantin Yurchenko, Jr on 9/11/24.
+//
+
+import Foundation
+
+/// Create a public chat channel.
+/// See [NIP-28](https://github.com/nostr-protocol/nips/blob/master/28.md#kind-40-create-channel).
+public class CreateChannelEvent: NostrEvent {
+    
+    public required init(from decoder: Decoder) throws {
+        try super.init(from: decoder)
+    }
+    
+    @available(*, unavailable, message: "This initializer is unavailable for this class.")
+    required init(kind: EventKind, content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+
+    @available(*, unavailable, message: "This initializer is unavailable for this class.")
+    required init(kind: EventKind, content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), pubkey: String) {
+        super.init(kind: kind, content: content, tags: tags, createdAt: createdAt, pubkey: pubkey)
+    }
+
+    @available(*, unavailable, message: "This initializer is unavailable for this class.")
+    override init(id: String, pubkey: String, createdAt: Int64, kind: EventKind, tags: [Tag], content: String, signature: String?) {
+        super.init(id: id, pubkey: pubkey, createdAt: createdAt, kind: kind, tags: tags, content: content, signature: signature)
+    }
+
+    init(content: String, tags: [Tag], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: Self.kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    class var kind: EventKind {
+        .channelCreation
+    }
+}
+
+public extension EventCreating {
+    func createChannelEvent(withContent content: String, signedBy keypair: Keypair) throws -> CreateChannelEvent {
+        return try CreateChannelEvent(content: content, tags: [], signedBy: keypair)
+    }
+}
+
--- a/Sources/NostrSDK/Events/Channels/CreateChannelMessageEvent.swift
+++ b/Sources/NostrSDK/Events/Channels/CreateChannelMessageEvent.swift
@@ -1,0 +1,52 @@
+//
+//  CreateChannelMessageEvent.swift
+//
+//
+//  Created by Konstantin Yurchenko, Jr on 9/11/24.
+//
+
+import Foundation
+
+/// Send a text message to a channel.
+/// See [NIP-28](https://github.com/nostr-protocol/nips/blob/master/28.md#kind-42-create-channel-message).
+public class CreateChannelMessageEvent: NostrEvent {
+    
+    public required init(from decoder: Decoder) throws {
+        try super.init(from: decoder)
+    }
+    
+    @available(*, unavailable, message: "This initializer is unavailable for this class.")
+    required init(kind: EventKind, content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+
+    @available(*, unavailable, message: "This initializer is unavailable for this class.")
+    required init(kind: EventKind, content: String, tags: [Tag] = [], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), pubkey: String) {
+        super.init(kind: kind, content: content, tags: tags, createdAt: createdAt, pubkey: pubkey)
+    }
+
+    @available(*, unavailable, message: "This initializer is unavailable for this class.")
+    override init(id: String, pubkey: String, createdAt: Int64, kind: EventKind, tags: [Tag], content: String, signature: String?) {
+        super.init(id: id, pubkey: pubkey, createdAt: createdAt, kind: kind, tags: tags, content: content, signature: signature)
+    }
+
+    init(content: String, tags: [Tag], createdAt: Int64 = Int64(Date.now.timeIntervalSince1970), signedBy keypair: Keypair) throws {
+        try super.init(kind: Self.kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    class var kind: EventKind {
+        .channelMessage
+    }
+}
+
+public extension EventCreating {
+    func createChannelMessageEvent(withContent content: String, eventId: String, relayUrl: String, signedBy keypair: Keypair) throws -> CreateChannelMessageEvent {
+        
+        var tags: [Tag] = [
+            Tag.pubkey(keypair.publicKey.hex),
+            Tag.event(eventId, otherParameters: [relayUrl, "root"]),
+        ]
+        
+        return try CreateChannelMessageEvent(content: content, tags: tags, signedBy: keypair)
+    }
+}
--- a/Sources/NostrSDK/Events/Channels/HideChannelMessageEvent.swift
+++ b/Sources/NostrSDK/Events/Channels/HideChannelMessageEvent.swift
@@ -1,0 +1,46 @@
+//
+//  HideChannelMessageEvent.swift
+//
+//
+//  Created by Konstantin Yurchenko, Jr on 9/11/24.
+//
+
+import Foundation
+
+/// User no longer wants to see a certain message.
+/// See [NIP-28](https://github.com/nostr-protocol/nips/blob/master/28.md#kind-43-hide-message).
+public class HideChannelMessageEvent: NostrEvent {
+    
+    public required init(from decoder: Decoder) throws {
+        try super.init(from: decoder)
+    }
+    
+    @available(*, unavailable, message: "This initializer is unavailable for this class.")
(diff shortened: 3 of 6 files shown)


Sources/NostrSDK/Events/Channels/CreateChannelEvent.swift
```diff
@@ -0,0 +1,47 @@
...
+        try super.init(kind: Self.kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    class var kind: EventKind {
+        .channelCreation
+    }
+}
+
+public extension EventCreating {
+    func createChannelEvent(withContent content: String, signedBy keypair: Keypair) throws -> CreateChannelEvent {
```
tyiu, line 43: I'm in the process of deprecating the `EventCreating` protocol actually, and moving towards a `Builder` pattern. The former is too rigid for building Nostr events, whereas the latter gives us a lot of flexibility.

See https://github.com/nostr-sdk/nostr-sdk-ios/pull/175/files#diff-c87e105b65b1d7c9512acf2ee8c7e3fc7582648811025c959a37b36d97403b67R170 as an example.

Also, it would be better to have an API where the developer can specify the basic channel metadata (name, about, picture and relays) rather than needing to enter in raw JSON in `content`.
https://github.com/nostr-protocol/nips/blob/master/28.md#kind-40-create-channel

Sources/NostrSDK/Events/Channels/CreateChannelMessageEvent.swift
```diff
@@ -0,0 +1,52 @@
...
+        try super.init(kind: Self.kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    class var kind: EventKind {
+        .channelMessage
+    }
+}
+
+public extension EventCreating {
+    func createChannelMessageEvent(withContent content: String, eventId: String, relayUrl: String, signedBy keypair: Keypair) throws -> CreateChannelMessageEvent {
```
tyiu, line 43: Same here. Again, look at how TextNoteEvent's Builder handles NIP-10 marker event tags. Maybe you could refactor or at least follow what we do in that code.

NIP-28 says that root or reply event tags are supported.

Sources/NostrSDK/Events/Channels/CreateChannelEvent.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+//
+//  Created by Konstantin Yurchenko, Jr on 9/11/24.
+//
+
+import Foundation
+
+/// Create a public chat channel.
+/// See [NIP-28](https://github.com/nostr-protocol/nips/blob/master/28.md#kind-40-create-channel).
+public class CreateChannelEvent: NostrEvent {
```
tyiu, line 12: It would be good to expose the basic channel metadata (name, about, picture and relays) properties.

Take a look at https://github.com/nostr-sdk/nostr-sdk-ios/blob/main/Sources/NostrSDK/Events/MetadataEvent.swift as it also does JSON encoding and decoding for the kind 0 event.

Sources/NostrSDK/Events/Channels/HideChannelMessageEvent.swift
```diff
@@ -0,0 +1,46 @@
...
+//
+//
+//  Created by Konstantin Yurchenko, Jr on 9/11/24.
+//
+
+import Foundation
+
+/// User no longer wants to see a certain message.
+/// See [NIP-28](https://github.com/nostr-protocol/nips/blob/master/28.md#kind-43-hide-message).
+public class HideChannelMessageEvent: NostrEvent {
```
tyiu, line 12: Same here for the `reason` metadata in the content field.

Sources/NostrSDK/Events/Channels/HideChannelMessageEvent.swift
```diff
@@ -0,0 +1,46 @@
...
+        try super.init(kind: Self.kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    class var kind: EventKind {
+        .channelHideMessage
+    }
+}
+
+public extension EventCreating {
+    func hideChannelMessageEvent(withContent content: String, signedBy keypair: Keypair) throws -> HideChannelMessageEvent {
```
tyiu, line 43: Same here.

Sources/NostrSDK/Events/Channels/MuteChannelUserEvent.swift
```diff
@@ -0,0 +1,46 @@
...
+//
+//
+//  Created by Konstantin Yurchenko, Jr on 9/11/24.
+//
+
+import Foundation
+
+/// User no longer wants to see messages from another user.
+/// See [NIP-28](https://github.com/nostr-protocol/nips/blob/master/28.md#kind-44-mute-user).
+public class MuteChannelUserEvent: NostrEvent {
```
tyiu, line 12: Same here about `reason`.

Sources/NostrSDK/Events/Channels/MuteChannelUserEvent.swift
```diff
@@ -0,0 +1,46 @@
...
+        try super.init(kind: Self.kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    class var kind: EventKind {
+        .channelMuteUser
+    }
+}
+
+public extension EventCreating {
+    func createMuteChannelUserEvent(withContent content: String, signedBy keypair: Keypair) throws -> MuteChannelUserEvent {
```
tyiu, line 43: Same here.

Sources/NostrSDK/Events/Channels/SetChannelMetadataEvent.swift
```diff
@@ -0,0 +1,46 @@
...
+//
+//
+//  Created by Konstantin Yurchenko, Jr on 9/11/24.
+//
+
+import Foundation
+
+/// Update a channel's public metadata.
+/// See [NIP-28](https://github.com/nostr-protocol/nips/blob/master/28.md#kind-41-set-channel-metadata).
+public class SetChannelMetadataEvent: NostrEvent {
```
tyiu, line 12: Same here about exposing the basic metadata fields as properties on this class.

Sources/NostrSDK/Events/Channels/SetChannelMetadataEvent.swift
```diff
@@ -0,0 +1,46 @@
...
+        try super.init(kind: Self.kind, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
+    }
+    
+    class var kind: EventKind {
+        .channelMetadata
+    }
+}
+
+public extension EventCreating {
+    func setChannelMetadataEvent(withContent content: String, signedBy keypair: Keypair) throws -> SetChannelMetadataEvent {
```
tyiu, line 43: Same here about the builder pattern and exposing an API to set the metadata fields instead of the raw content field.

tyiu (changes requested):
Thanks for the contribution! I left a few comments around supporting properties of each kind in a first-class way, and using a `Builder` protocol instead of continuing with the soon-to-be-deprecated `EventCreating` protocol.

Also, there seems to be a few SwiftLint errors that should be addressed.

Sources/NostrSDK/Events/Channels/CreateChannelEvent.swift
masterial: Hi, @tyiu, thanks for your time and looking over my commit! 

Yes, I am on board with the later. The interface in this PR needs further definition and polish. 
I created this PR because I am using your SDK with my project (SwiftUI/Swift 5). Thanks for putting it together!
So far, I was able to implement and successfully field test Encrypted Message, Group Chat and Key Management features. 
As far as, my PR goes I will clean it up soon for the linter. I am thinking doing it in parallel while I’m working on replies for my chat application. 

I don't quite understand what you mean by ‘deprecating EventCreating protocol’. I do use it a lot but I did find it somewhat rigid.
I also think there needs to be some better way to manage subscriptions. I do have them on about 5 Views on my app. Yikes...! 🦟

https://github.com/SkatePay/skatepay/blob/main/SkateConnect/SkateConnect/Views/Skatepark/Lobby/SpotFeed/ChannelFeed.swift#L145

Regards!

Sources/NostrSDK/Events/Channels/CreateChannelMessageEvent.swift
masterial: Yes, I will take care of this when I am working on replies in my app.

Sources/NostrSDK/Events/Channels/CreateChannelEvent.swift
masterial: Should I make a MetadataChannel class here or what?

Sources/NostrSDK/Events/Channels/MuteChannelUserEvent.swift
masterial: Yes, implementing this feature soon too.

Sources/NostrSDK/Events/Channels/CreateChannelEvent.swift
tyiu: In a separate file, you could create a `ChannelMetadata` struct with the properties. Then, you could create a `ChannelMetadataInterpreting` protocol that decodes the JSON.

Kinds 40 `CreateChannelEvent` and kind 41 `SetChannelMetadataEvent` could extend from `ChannelMetadataInterpreting`.

Something like this, it'll be pretty similar to what we do with kind 0:
```swift
public struct ChannelMetadata: Codable {
    public let name: String?
    public let about: String?
    ...

    enum CodingKeys: String, CodingKey {
        ...
    }
}

public protocol ChannelMetadataInterpreting: NostrEvent {}
public extension ChannelMetadataInterpreting {
    public var channelMetadata: ChannelMetadata {
        guard let data = content.data(using: .utf8) else {
            return nil
        }
        return try? JSONDecoder().decode(ChannelMetadata.self, from: data)
    }
}
```

Sources/NostrSDK/Events/Channels/CreateChannelEvent.swift
tyiu: When I say deprecating the `EventCreating` protocol in favor of the `Builder` protocol, I mean it would look something like this when creating this event:

```swift
try CreateChannelEvent.Builder()
    .channelMetadata(channelMetadata, merging: rawChannelMetadata)
    .build(signedBy: keypair)
```

Take a look at what we do here for kind 0, as an example:
https://github.com/nostr-sdk/nostr-sdk-ios/blob/8b40eb98a4268255b8980bddddf40125b0d32637/Sources/NostrSDK/Events/MetadataEvent.swift#L161-L192

tyiu: @masterial Turns out the SwiftLint errors were on the main branch, due to how CI pulls in the latest SwiftLint version and it had some breaking changes in what it validates. I just fixed the issue on the main branch so the lint errors should go away after you pull it into your branch.

masterial: Merged, thanks for the heads up, @tyiu!
I plan to address other issues you've highlighted very soon

masterial: Hi Terry, @tyiu 

I hope you're doing well. I'm reaching out because I'm delving into the world of Nostr SDK implementation, and I've encountered a few hurdles that I believe you might help me navigate:

Relay Connections: I'm experiencing issues with maintaining persistent relay connections. There seems to be a timeout problem. Could you guide me on how to best handle this? Specifically, what's the recommended approach for reconnecting relays when they drop?
Relay Pool Lifecycle: I'm unsure about managing the lifecycle of the relay pool. What are the best practices here? Should I be constantly monitoring and refreshing connections, or is there a more efficient way?
Debugging Events: Currently, there's a lot of event debugging happening in my setup which I'd like to minimize. Is there a way to disable or manage this more efficiently?

On another note, I'm excited to share that my app, [SkateConnect](https://support.skatepark.chat/) (also known as SkatePay), has just been released on the App Store. We're now focusing on enhancing its robustness, particularly with connection persistence. Given your expertise, I'd love to discuss potential collaborations or advisory roles concerning Nostr integration. If you're open to it, I'd appreciate discussing this further, perhaps exploring how we might make Nostr work more seamlessly within SkateConnect. (We can talk over nostr too, not here).

Looking forward to your thoughts and any advice you can offer!

Best regards!

PS. Still got some work to do before I polish this PR.

Sources/NostrSDK/Events/Channels/CreateChannelMessageEvent.swift
tyiu: FYI I'm working on [NIP-17 - Private Direct Messages](https://github.com/nostr-protocol/nips/blob/master/17.md) which requires threading as well, so I ended up refactoring the logic for building events with NIP-10 tags. It hasn't been merged yet -- still working on my branch but it'll be ready soon. Hope you haven't gotten to it yet, don't want to make your changes redundant.

masterial: Great! Thank you so much.

tyiu: Merged: https://github.com/nostr-sdk/nostr-sdk-ios/pull/186

PR #184, opened 2024-10-17
Add support for receiver aliases for gift wrap creation to avoid exposing identities and improve test coverage

tyiu:
This should be a backwards compatible change.

### DIFF
--- a/Sources/NostrSDK/Events/GiftWrap/GiftWrapEvent.swift
+++ b/Sources/NostrSDK/Events/GiftWrap/GiftWrapEvent.swift
@@ -87,36 +87,38 @@ public extension EventCreating {
     /// Creates a ``GiftWrapEvent`` that takes a rumor, an unsigned ``NostrEvent``, and seals it in a signed ``SealEvent``, and then wraps that seal encrypted in the content of the gift wrap.
     ///
     /// - Parameters:
-    ///   - withRumor: a ``NostrEvent`` that is not signed.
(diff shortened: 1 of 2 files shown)


Sources/NostrSDK/Events/GiftWrap/GiftWrapEvent.swift
```diff
@@ -87,36 +87,38 @@ public extension EventCreating {
...
     ///   - tags: the list of tags.
     ///   - createdAt: the creation timestamp of the seal. Note that this timestamp SHOULD be tweaked to thwart time-analysis attacks. Note that some relays don't serve events dated in the future, so all timestamps SHOULD be in the past. By default, if `createdAt` is not provided, a random timestamp within 2 days in the past will be chosen.
-    ///   - keypair: The real ``Keypair`` to sign the seal with. Note that a different random one-time use key is used to sign the gift wrap.
     func giftWrap(
         withSeal seal: SealEvent,
         toRecipient recipient: PublicKey,
+        recipientAlias: PublicKey? = nil,
         tags: [Tag] = [],
-        createdAt: Int64 = Int64(Date.now.timeIntervalSince1970 - TimeInterval.random(in: 0...172800)),
-        signedBy keypair: Keypair
```
tyiu, line 119: This parameter was not being used and does not make sense in the context of this function because the seal event already exists -- there's nothing to sign with it and the gift wrap is signed with a one-time-use random keypair.

PR #186, opened 2024-10-18
Refactor threaded event tag building and interpreting logic so non-kind-1 events can reuse them

tyiu:
For example, [NIP-17 - Private Direct Message](https://github.com/nostr-protocol/nips/blob/master/17.md) uses threading, so this PR unblocks that work from happening in the near future.


Sources/NostrSDK/Events/Tags/EventTag.swift
```diff
@@ -150,11 +150,160 @@ public struct EventTag: RelayProviding, RelayURLValidating, Equatable {
     }
 }
 
-public protocol EventTagInterpreting: NostrEvent {}
```
tyiu, line 153: Unused.

Sources/NostrSDK/Events/Tags/EventTag.swift
```diff
@@ -150,11 +150,160 @@ public struct EventTag: RelayProviding, RelayURLValidating, Equatable {
...
+            insertTags(try EventTag(eventId: repliedEvent.id, relayURL: validatedRelayURL, marker: .root, pubkey: repliedEvent.pubkey).tag, at: 0)
+        }
+
+        // When replying to a text event E, the reply event's "p" tags should contain all of E's "p" tags as well as the "pubkey" of the event being replied to.
+        // Example: Given a text event authored by a1 with "p" tags [p1, p2, p3] then the "p" tags of the reply should be [a1, p1, p2, p3] in no particular order.
+        appendTags(contentsOf: repliedEvent.tags.filter { $0.name == TagName.pubkey.rawValue })
+
+        // Add the author "p" tag if it was not already added.
+        if !tags.contains(where: { $0.name == TagName.pubkey.rawValue && $0.value == repliedEvent.pubkey }) {
+            if let validatedRelayURL {
```
tyiu, line 277: This specific logic around adding a relay to the p-tag is new. Which is why `TextNoteEventTests` was updated to match the assertion. All the other logic here was copied from `TextNoteEvent`.

Sources/NostrSDK/Events/TextNoteEvent.swift
```diff
@@ -37,87 +37,25 @@ public final class TextNoteEvent: NostrEvent, CustomEmojiInterpreting {
...
-        // then we can make a reasonable assumption that the client that created the event does not use
-        // deprecated positional event tags, so there is no root event tag.
-        guard eventTags.allSatisfy({ $0.marker == nil }) else {
-            return nil
-        }
-
-        // NIP-10 states that the first event tag is the root.
-        return eventTags.first
+    /// Events mentioned in the note content.
+    @available(*, deprecated, message: "Deprecated in favor of referencedEventIds. Mention is an overloaded term from NIP-10.")
```
tyiu, line 51: [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md#tags) says
```
- The `e` tag, used to refer to an event: `["e", <32-bytes lowercase hex of the id of another event>, <recommended relay URL, optional>]`
```

Since this line uses the word `refer`, I've called the new property `referencedEventIds`.

Sources/NostrSDK/Events/TextNoteEvent.swift
```diff
@@ -36,88 +36,26 @@ public final class TextNoteEvent: NostrEvent, CustomEmojiInterpreting {
         try super.init(kind: .textNote, content: content, tags: tags, createdAt: createdAt, signedBy: keypair)
     }
 
-    /// Pubkeys mentioned in the note content.
+    /// Pubkeys referenced in the note content.
+    @available(*, deprecated, message: "Deprecated in favor of referencedPubkeys. Mention is an overloaded term from NIP-10.")
```
tyiu, line 40: [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md#tags) says
```
- The `p` tag, used to refer to another user: `["p", <32-bytes lowercase hex of a pubkey>, <recommended relay URL, optional>]`
```

Since this line uses the word `refer`, I've called the new property `referencedPubkeys`.

PR #195, opened 2024-11-03
Add support for NIP-32 Labeling

tyiu:
https://github.com/nostr-protocol/nips/blob/master/32.md

### DIFF
--- a/Sources/NostrSDK/Events/NostrEvent.swift
+++ b/Sources/NostrSDK/Events/NostrEvent.swift
@@ -10,7 +10,7 @@ import Foundation
 /// A structure that describes a Nostr event.
 ///
 /// > Note: [NIP-01 Specification](https://github.com/nostr-protocol/nips/blob/master/01.md#events-and-signatures)
-public class NostrEvent: Codable, Equatable, Hashable {
+public class NostrEvent: Codable, Equatable, Hashable, LabelTagInterpreting {
     public static func == (lhs: NostrEvent, rhs: NostrEvent) -> Bool {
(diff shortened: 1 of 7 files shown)


Sources/NostrSDK/Events/NostrEvent.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 /// A structure that describes a Nostr event.
 ///
 /// > Note: [NIP-01 Specification](https://github.com/nostr-protocol/nips/blob/master/01.md#events-and-signatures)
-public class NostrEvent: Codable, Equatable, Hashable {
+public class NostrEvent: Codable, Equatable, Hashable, LabelTagInterpreting {
```
tyiu, line 13: Although labels can exist on any NostrEvent, I've chosen to separate that part of the class out into a separate protocol/extension so that it's easier to maintain, and provides a logical separation.

Sources/NostrSDK/Events/NostrEvent.swift
```diff
@@ -318,7 +318,7 @@ public protocol NostrEventBuilding {
 
 public extension NostrEvent {
     /// Builder of a ``NostrEvent`` of type `T`.
-    class Builder<T: NostrEvent>: NostrEventBuilding {
+    class Builder<T: NostrEvent>: NostrEventBuilding, LabelBuilding {
```
tyiu, line 321: Although labels can exist on any NostrEvent, I've chosen to separate that part of the class out into a separate protocol/extension so that it's easier to maintain, and provides a logical separation.

PR #197, opened 2024-11-03
Refactor out common non-NIP-01 tags out of NostrEvent file into separ…

tyiu:
…ate protocols for better code separation

### DIFF
--- a/Sources/NostrSDK/Events/NostrEvent.swift
+++ b/Sources/NostrSDK/Events/NostrEvent.swift
@@ -10,7 +10,7 @@ import Foundation
 /// A structure that describes a Nostr event.
 ///
 /// > Note: [NIP-01 Specification](https://github.com/nostr-protocol/nips/blob/master/01.md#events-and-signatures)
-public class NostrEvent: Codable, Equatable, Hashable, ContentWarningTagInterpreting, LabelTagInterpreting {
+public class NostrEvent: Codable, Equatable, Hashable, AlternativeSummaryTagInterpreting, ContentWarningTagInterpreting, ExpirationTagInterpreting, LabelTagInterpreting {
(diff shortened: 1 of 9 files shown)


Sources/NostrSDK/Events/NostrEvent.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 /// A structure that describes a Nostr event.
 ///
 /// > Note: [NIP-01 Specification](https://github.com/nostr-protocol/nips/blob/master/01.md#events-and-signatures)
-public class NostrEvent: Codable, Equatable, Hashable, ContentWarningTagInterpreting, LabelTagInterpreting {
+public class NostrEvent: Codable, Equatable, Hashable, AlternativeSummaryTagInterpreting, ContentWarningTagInterpreting, ExpirationTagInterpreting, LabelTagInterpreting {
```
tyiu, line 13: Making this change allows us to keep the base NostrEvent class clean for only the NIP-01 implementation. Any other NIPs that apply to all event kinds can be introduced as a separate protocol/extension that NostrEvent extends from. It'll let us maintain the code much easier.

Sources/NostrSDK/Events/NostrEvent.swift
```diff
@@ -318,7 +291,7 @@ public protocol NostrEventBuilding {
 
 public extension NostrEvent {
     /// Builder of a ``NostrEvent`` of type `T`.
-    class Builder<T: NostrEvent>: NostrEventBuilding, ContentWarningTagBuilding, LabelBuilding {
+    class Builder<T: NostrEvent>: NostrEventBuilding, AlternativeSummaryTagBuilding, ContentWarningTagBuilding, ExpirationTagBuilding, LabelTagBuilding {
```
tyiu, line 294: Same here.

PR #204, opened 2025-01-04
Add nip44 encrypt/decrypt example to demo project

fabianfabian:
I added an example for encrypting/decrypting with nip44 to the demo project, I was using this myself to help me understand but might be useful for others too.

I copied the style from LegacyDirectMessageDemoView but limited it to encrypt/decrypt, no publishing to relays.


tyiu (approved):
Looks great. I confirmed that it works by:
1. Encrypting a message with public key A, private key B
2. Decrypting the encrypted message with public key A and private key B
3. Decrypting the encrypted message with public key B and private key A

Thanks for the contribution!

PR #216, opened 2026-01-08
Fix bytes conflict 0.3.0

tkhumush:
This should be a simple fix. 

My could not build without this fix. But now it does and working fine. 

if this is not correct, I would appreciate the feedback, because I dont want to leave my app on it if it creates other problems I not seeing immediately.


tyiu (approved):
Looks good. Thanks!

tyiu:
The macOS 13 test runner is deprecated so I'll need to remove those unit tests so that I can merge this PR. I'll do that later today.

